PR #49, opened 2020-08-23
Gender and date of birth to the user added.

pacificsun:
Hi, Ian here is the PR Please review it and any feedback will be appreciated! :)

### DIFF
--- a/integration-testing/tests/users/properties.test.js
+++ b/integration-testing/tests/users/properties.test.js
@@ -47,14 +47,18 @@ describe('Read and write properties our our own profile', () => {
     expect(resp.data.user.privacyStatus).toBe('PUBLIC')
   })
 
-  test('fullName and bio', async () => {
+  test('fullName, bio, gender and dob,',  async () => {
     const bio = "truckin'"
     const fullName = 'Hunter S.'
+    const dob="1995-05-05",
+    const gender = "Male"
     const {client, userId} = await loginCache.getCleanLogin()
 
     let resp = await client.query({query: queries.user, variables: {userId}})
     expect(resp.data.user.bio).toBeNull()
     expect(resp.data.user.fullName).toBeNull()
+    expect(resp.data.user.dob).toBeNull()
+    expect(resp.data.user.gender).toBeNull()
 
     // set to some custom values
     resp = await client.mutate({mutation: mutations.setUserDetails, variables: {bio, fullName}})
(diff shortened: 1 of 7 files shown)


integration-testing/tests/users/properties.test.js
```diff
@@ -88,6 +110,12 @@ test('Disabled user cannot setUserDetails', async () => {
   await expect(client.mutate({mutation: mutations.setUserDetails, variables: {bio: 'a dog'}})).rejects.toThrow(
     /ClientError: User .* is not ACTIVE/,
   )
+  await expect(client.mutate({mutation: mutations.setUserDetails, variables: {dob: '1995-05-05'}})).rejects.toThrow(
+    /ClientError: User .* is not ACTIVE/,
+  )
+  await expect(client.mutate({mutation: mutations.setUserDetails, variables: {gender: 'Male'}})).rejects.toThrow(
+    /ClientError: User .* is not ACTIVE/,
+  )
```
mfogel, line 118: I wouldn't bother to test that `dob` and `gender` throw errors when using `setUserDetails()` here. The test is focused on testing if disabling users works, so I think it's sufficient to test that by trying to set one random property to something... that random property could be `bio` or `dob` or `gender`, doesn't really matter which. But all three is overkill, in my opinion.

real-main/schema.graphql
```diff
@@ -32,7 +32,11 @@ type Query {
   user(userId: ID!): User
```
mfogel, line 32: It looks like your auto-formatter did a lot to this file. The new format is not necessarily good or bad, but it doesn't belong in this PR. Better to separate re-formats of files into separate PR's.

real-main/schema.graphql
```diff
@@ -98,18 +123,20 @@ type Mutation {
...
-    viewCountsHidden: Boolean,
-    commentsDisabled: Boolean,
-    likesDisabled: Boolean,
-    sharingDisabled: Boolean,
-    verificationHidden: Boolean,
-    languageCode: String,
+    username: String
+    fullName: String
+    bio: String
+    dob: String
```
mfogel, line 129: Better to use AWS's custom scalar date type `AWSDate` here: https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html

real-main/schema.graphql
```diff
@@ -98,18 +123,20 @@ type Mutation {
...
+    gender: String
```
mfogel, line 130: Better to type this a as new custom graphql enum. Possible values would be at least `M` and `F`, and any others we want to add.

pacificsun:
Thank you for the feedback and i will try to improve those points :)

PR #50, opened 2020-08-23
DOB,Gender added to user flow and Dating Matches and Potential matches

manish2kcoder:
1. DOB and Gender added to user flow 
2. Dating Matches and Potential matches

### DIFF
--- a/real-main/app/handlers/appsync/handlers.py
+++ b/real-main/app/handlers/appsync/handlers.py
@@ -23,8 +23,8 @@ from app.models.post.enums import PostStatus, PostType
 from app.models.post.exceptions import PostException
 from app.models.user.enums import UserStatus
 from app.models.user.exceptions import UserException
-from app.utils import image_size
 from app.models.match.manager import MatchManager
+from app.utils import image_size
 
 from .. import xray
 from . import routes
--- a/real-main/app/handlers/appsync/handlers.py
+++ b/real-main/app/handlers/appsync/handlers.py
@@ -1,6 +1,7 @@
-import pendulum
 import logging
 import os
+
+import pendulum
 
 from app import clients, models
 from app.mixins.flag.enums import FlagStatus
@@ -24,10 +25,9 @@ from app.models.post.exceptions import PostException
 from app.models.user.enums import UserStatus
 from app.models.user.exceptions import UserException
 from app.utils import image_size
-
-from .. import xray
 from . import routes
 from .exceptions import ClientException
+from .. import xray
 
 S3_UPLOADS_BUCKET = os.environ.get('S3_UPLOADS_BUCKET')
 S3_PLACEHOLDER_PHOTOS_BUCKET = os.environ.get('S3_PLACEHOLDER_PHOTOS_BUCKET')
--- a/real-main/app/handlers/appsync/handlers.py
+++ b/real-main/app/handlers/appsync/handlers.py
@@ -1220,7 +1220,7 @@ def lambda_server_error(caller_user_id, arguments, source, context):
 
 
 @routes.register('Mutation.potentialMatchesPageCount')
-def get_potential_matches_count(caller_user_id, arguments, source, context):
+def get_potential_matches__page_count(caller_user_id, arguments, source, context):
     try:
         potential_matches = match_manager.get_potential_matches(caller_user_id)
     except UserException as err:
@@ -1232,7 +1232,7 @@ def get_potential_matches_count(caller_user_id, arguments, source, context):
 
 
 @routes.register('Mutation.matchesPageCount')
-def get_matches(caller_user_id, arguments, source, context):
+def get_matches_page_count(caller_user_id, arguments, source, context):
     try:
         match_list = match_manager.get_common_likes(caller_user_id)
     except UserException as err:
--- a/real-main/app/handlers/appsync/handlers.py
+++ b/real-main/app/handlers/appsync/handlers.py
@@ -24,6 +24,7 @@ from app.models.post.exceptions import PostException
 from app.models.user.enums import UserStatus
 from app.models.user.exceptions import UserException
 from app.utils import image_size
+from app.models import LikeManager, MatchManager
 
 from .. import xray
 from . import routes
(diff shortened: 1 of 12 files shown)


real-main/app/handlers/appsync/handlers.py
```diff
@@ -1190,3 +1198,79 @@ def lambda_client_error(caller_user_id, arguments, source, context):
...
+    try:
+        match_list = match_manager.get_common_likes(caller_user_id)
+    except UserException as err:
+        raise ClientException(str(err))
+    if page_num:
+        start_indx = page_num * MATCH_PAGE_SIZE
+        end_indx = start_indx + MATCH_PAGE_SIZE
+        if start_indx >= len(match_list) and end_indx <= len(match_list):
+            return match_list[start_indx:end_indx]
+    return match_list  # Will need to serialize for http
```
mfogel, line 1276: Nice job adding some python lambda handlers here for querying matches.

If I were implementing this myself, I would actually do all these reads in VTL mapping templates, not in lambda functions. The upsides to this are 1) cost - AWS doesn't charge for 'running' VTL templates, but it certainly does for lambda and 2) response times are better. The downside to this is the learning curve. Coding in VTL is like wandering through a minefield of unintuitive gotchas.

With respect to the pagination, the pattern we are following in this project follows from AWS's built-in pagination AppSync dynamodb pagination: https://docs.aws.amazon.com/appsync/latest/devguide/resolver-mapping-template-reference-dynamodb.html#aws-appsync-resolver-mapping-template-reference-dynamodb-query. So Paginated graphql responses have this form, for example: https://github.com/real-social-media/backend/blob/v2020-08-20/real-main/schema.graphql#L798-L801 The `nextToken` is an obscured dynamo pagination token.

When/if we do need to implement a lambda function that return a paginated response, here's a python function that generates a `nextToken` given a dynamo pagination token: https://github.com/real-social-media/backend/blob/v2020-08-20/real-main/app/clients/dynamo.py#L148-L151

real-main/app/models/match/dynamo.py
```diff
@@ -0,0 +1,35 @@
...
+        query_kwargs = {
+            'KeyConditionExpression': Key('gsiA1PartitionKey').eq(f'like/{liked_by_user_id}'),
+            'IndexName': 'GSI-A1',
+        }
+        liked_users = self.client.generate_all_query(query_kwargs)
+        matches = []
+        for liked_user in liked_users:
+            if liked_user["likedByUserId"] == liked_by_user_id:
+                matches.append(liked_user)
+        return matches
```
mfogel, line 35: I'm not sure using a key of `{'partitionKey': f'post/{post_id}', 'sortKey': f'like/{liked_by_user_id}'}` is best. I might have done something like  `{'partitionKey': f'user/{user_id}', 'sortKey': f'match/{other_user_id}'}`

real-main/app/models/match/exceptions.py
```diff
@@ -0,0 +1,11 @@
+class MatchException(Exception):
+    pass
+
+
+class NoMatchFound(MatchException):
+    def __init__(self, user_id, post_id):
```
mfogel, line 6: argument `post_id` is unused.

real-main/app/models/match/manager.py
```diff
@@ -0,0 +1,38 @@
...
+class MatchManager:
+    def __init__(self, clients, managers=None):
+        managers = managers or {}
+        managers['match'] = self
+        self.follower_manager = managers.get('follower') or models.FollowerManager(clients, managers=managers)
+        self.clients = clients
+        if 'dynamo' in clients:
+            self.dynamo = MatchDynamo(clients['dynamo'])
+
+    """ All users not in contact list and with mutual friends should ideally be potential matches """
```
mfogel, line 19: Is this a docstring? If so, it should go as the first line in a method.

real-main/app/models/match/model.py
```diff
@@ -0,0 +1,11 @@
+import logging
+
+logger = logging.getLogger()
+
+
+class Match:
+    def __init__(
+        self, user_id, match_dynamo,
+    ):
```
mfogel, line 9: formatting - no need for arguments to be on their own line

real-main/app/models/match/model.py
```diff
@@ -0,0 +1,11 @@
...
+        self.dynamo = match_dynamo
+        self.liked_by_user_id = user_id
```
mfogel, line 11: variable naming - is there any liking going on here?

real-main/app_tests/models/match/test_dynamo.py
```diff
@@ -0,0 +1,99 @@
...
+
+from app.models.like.enums import LikeStatus
+from app.models.match.dynamo import MatchDynamo
+
+
+def match_dynamo(dynamo_client):
+    yield MatchDynamo(dynamo_client)
+
+
+@pytest.fixture
```
mfogel, line 12: this appears to be a test, not a fixture - so no need for the decorator.

real-main/app_tests/models/match/test_dynamo.py
```diff
@@ -0,0 +1,99 @@
...
+def test_get_common_likes(like_dynamo, match_dynamo):
+    generate_test_data(like_dynamo)
+    matches = match_dynamo.get_common_likes('lbuid')
+    assert matches is None
+    assert len(matches) == 1
+
+
+def generate_test_data(like_dynamo):
```
mfogel, line 31: With pytest, instead of writing functions like this to generate data to test on you write fixtures. If you haven't already, I encourage you to check them out - super powerful https://docs.pytest.org/en/stable/fixture.html

real-main/app_tests/models/user/test_dynamo.py
```diff
@@ -440,6 +440,36 @@ def test_set_user_privacy_status(user_dynamo):
...
+    gender = 'M'
+    user_item = user_dynamo.add_user(user_id, username, gender=gender)
+    assert user_item['gender'] == 'M'
+
+
+def test_user_empty_gender(user_dynamo):
+    user_id = 'my-user-id'
+    username = 'my-username'
+    user_item = user_dynamo.add_user(user_id, username)
+    assert 'gender' not in user_item
```
mfogel, line 470: Nice job on these tests.

real-main/schema.graphql
```diff
@@ -110,7 +110,9 @@ type Mutation {
     sharingDisabled: Boolean,
     verificationHidden: Boolean,
     languageCode: String,
-    themeCode: String
+    themeCode: String,
+    gender: String,
+    dob: String,
```
mfogel, line 115: It would be better to type `dob` as a `AWSDate` rather than a string https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html

And it would be better to type `gender` as a custom graphql enum with types `M` and `F` and what ever else we would like to add, rather than a string.

PR #51, opened 2020-08-24
Add 2 user fields update unittest with pytest and integration test.

YasinFatullayev:
- Add birthday&gender fields to user.
- Update unittest with pytest and integration test.

### DIFF
--- a/integration-testing/tests/users/create-user/cognito-only.test.js
+++ b/integration-testing/tests/users/create-user/cognito-only.test.js
@@ -185,17 +185,17 @@ describe('cognito-only user', () => {
       const username = cognito.generateUsername()
       const fullName = 'Yasin Fatullayev'
       const birthday = '1900-01-01'
-      const gender = 'Man'
+      const gender = 'Male'
       let variables = {username, fullName, birthday, gender}
       let resp = await client.mutate({mutation: mutations.createCognitoOnlyUser, variables})
 
-      expect(resp['errors']).toBeUndefined()
-      expect(resp['data']['createCognitoOnlyUser']['userId']).toBe(userId)
-      expect(resp['data']['createCognitoOnlyUser']['username']).toBe(username)
-      expect(resp['data']['createCognitoOnlyUser']['email']).toBe(email)
-      expect(resp['data']['createCognitoOnlyUser']['fullName']).toBe(fullName)
-      expect(resp['data']['createCognitoOnlyUser']['birthday']).toBe(birthday)
-      expect(resp['data']['createCognitoOnlyUser']['gender']).toBe(gender)
+      expect(resp.errors).toBeUndefined()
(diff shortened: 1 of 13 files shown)


integration-testing/tests/users/create-user/cognito-only.test.js
```diff
@@ -180,5 +180,22 @@ describe('cognito-only user', () => {
...
+      let variables = {username, fullName, birthday, gender}
+      let resp = await client.mutate({mutation: mutations.createCognitoOnlyUser, variables})
+
+      expect(resp['errors']).toBeUndefined()
+      expect(resp['data']['createCognitoOnlyUser']['userId']).toBe(userId)
+      expect(resp['data']['createCognitoOnlyUser']['username']).toBe(username)
+      expect(resp['data']['createCognitoOnlyUser']['email']).toBe(email)
+      expect(resp['data']['createCognitoOnlyUser']['fullName']).toBe(fullName)
+      expect(resp['data']['createCognitoOnlyUser']['birthday']).toBe(birthday)
+      expect(resp['data']['createCognitoOnlyUser']['gender']).toBe(gender)
```
mfogel, line 198: One place modern javascript beats python in my opinion: You can write these as using `.` to access object properties instead of `['']`. Ex: `expect(resp.data'.createCognitoOnlyUser.gender).toBe(gender)`

real-main/schema.graphql
```diff
@@ -64,7 +64,7 @@ type Query {
 
 type Mutation {
   # Set up a new user account
-  createCognitoOnlyUser(username: String!, fullName: String): User
+  createCognitoOnlyUser(username: String!, fullName: String, birthday: String, gender: String): User
```
mfogel, line 67: The graphql typing could tightened a bit here - which I would recommend.

Instead of typing `birthday` as a `String`, AWS provides a custom scalar type `AWSDate` https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html which would work nicely here.

And instead of typing `gender` as a `String`, I would make a customer graphql enum with options of `M` and `F` and whatever other gender options we want to add.

integration-testing/tests/users/create-user/cognito-only.test.js
YasinFatullayev: Resolved.
Please check.

real-main/schema.graphql
YasinFatullayev: Resolved.
Please check.

mfogel:
Thank you for addressing my comments @YasinFatullayev . Excellent work.

mfogel:
@YasinFatullayev when you have time, could you rebase this against `master`, and the re-submit the PR for merge into `master`? Nice work 😊

mfogel:
Closing in favor of #55

PR #52, opened 2020-08-24
Add gender and birthDate field, write tests

WonderCode22:


real-main/app/models/user/dynamo/base.py
```diff
@@ -37,8 +37,18 @@ def delete_user(self, user_id):
         return self.client.delete_item(self.pk(user_id))
 
     def add_user(
-        self, user_id, username, full_name=None, email=None, phone=None, placeholder_photo_code=None, now=None
+        self,
+        user_id,
+        username,
+        full_name=None,
+        gender=None,
+        birthDate=None,
```
mfogel, line 45: To more closely follow python's conventions, I would name this kwarg `birth_date` rather than `birthDate`

real-main/mapping-templates/User.birthDate.request.vtl
```diff
@@ -0,0 +1,9 @@
+## birthDate is private to the user themselves
+#if ($ctx.source.userId != $ctx.identity.cognitoIdentityId)
+  #return ()
+#end
+
+{
+  "version": "2018-05-29",
+  "payload": $util.toJson($ctx.source.birthDate)
+}
```
mfogel, line 9: Nice job writing some VTL!

real-main/schema.graphql
```diff
@@ -361,6 +363,8 @@ type User {
   # private to user themselves only
   email: AWSEmail
   phoneNumber: AWSPhone
+  gender: String
+  birthDate: AWSDateTime
```
mfogel, line 367: Nice job using something other than `String` as the type for `birthDate` here :). That being said, I probably would have used an `AWSDate` rather than a `AWSDateTime`.

For `gender`, I would probably type it as a custom graphql enum with possible values `M` and `F` and whatever other genders we want to support.

Then finally, I think the types required in the arguments to `setUserDetails()` for `birthDate` and `gender` should match those in the `User` type.

PR #53, opened 2020-08-24
add two fields(date_of_birth and gender) and swiping api

heroclass728:
- Add date_of_birth and gender fields into user
- Update integration test
- Swiping api

### DIFF
--- a/integration-testing/tests/users/properties.test.js
+++ b/integration-testing/tests/users/properties.test.js
@@ -47,14 +47,18 @@ describe('Read and write properties our our own profile', () => {
     expect(resp.data.user.privacyStatus).toBe('PUBLIC')
   })
 
-  test('fullName and bio', async () => {
+  test('fullName and bio, date_of_birth and gender', async () => {
     const bio = "truckin'"
     const fullName = 'Hunter S.'
+    const date_of_birth='2020-08-20'
+    const gender = "man"
     const {client, userId} = await loginCache.getCleanLogin()
 
     let resp = await client.query({query: queries.user, variables: {userId}})
     expect(resp.data.user.bio).toBeNull()
     expect(resp.data.user.fullName).toBeNull()
+    expect(resp.data.user.date_of_birth).toBeNull()
+    expect(resp.data.user.gender).toBeNull()
 
     // set to some custom values
     resp = await client.mutate({mutation: mutations.setUserDetails, variables: {bio, fullName}})
@@ -64,6 +68,13 @@ describe('Read and write properties our our own profile', () => {
     resp = await client.query({query: queries.user, variables: {userId}})
     expect(resp.data.user.bio).toBe(bio)
     expect(resp.data.user.fullName).toBe(fullName)
+    expect(resp.data.user.date_of_birth).toBeNull()
+    expect(resp.data.user.gender).toBeNull()
+
+    // set to date_of_birth, gender
+    resp = await client.mutate({mutation: mutations.setUserDetails, variables: {date_of_birth, gender}})
+    expect(resp.data.setUserDetails.date_of_birth).toBe(date_of_birth)
+    expect(resp.data.setUserDetails.gender).toBe(gender)
 
     // clear out the custom values
     resp = await client.mutate({mutation: mutations.setUserDetails, variables: {bio: '', fullName: ''}})
@@ -73,6 +84,15 @@ describe('Read and write properties our our own profile', () => {
     resp = await client.query({query: queries.user, variables: {userId}})
     expect(resp.data.user.bio).toBeNull()
     expect(resp.data.user.fullName).toBeNull()
+
+    // clear out date_of_birth, gender
+    resp = await client.mutate({mutation: mutations.setUserDetails, variables: {date_of_birth: '', gender: ''}})
+    expect(resp.data.setUserDetails.date_of_birth).toBeNull()
+    expect(resp.data.setUserDetails.gender).toBeNull()
+
+    resp = await client.query({query: queries.user, variables: {userId}})
+    expect(resp.data.user.date_of_birth).toBeNull()
+    expect(resp.data.user.gender).toBeNull()
   })
 })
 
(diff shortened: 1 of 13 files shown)


integration-testing/tests/users/properties.test.js
```diff
@@ -160,6 +180,13 @@ test('Various photoPostId failures', async () => {
   expect(resp.data.addPost.postId).toBe(postId)
   expect(resp.data.addPost.postStatus).toBe('COMPLETED')
   expect(resp.data.addPost.postType).toBe('IMAGE')
+  expect(resp.data.addPost.isVerified).toBe(false)
+
+  // verify can't set our profile photo using non-verified post
+  variables = {photoPostId: postId}
+  await expect(ourClient.mutate({mutation: mutations.setUserDetails, variables})).rejects.toThrow(
+    /ClientError: .*is not verified/,
+  )
```
mfogel, line 189: Kinda confused by this staza. It looks ok, but not sure it belongs in this PR.

real-main/app/handlers/appsync/handlers.py
```diff
@@ -117,6 +123,20 @@ def create_facebook_user(caller_user_id, arguments, source, context):
...
+    google_id_token = arguments['googleIdToken']
+    try:
+        user = user_manager.create_federated_user(
+            'google', caller_user_id, username, google_id_token, full_name=full_name
+        )
+    except UserException as err:
+        raise ClientException(str(err))
+    return user.serialize(caller_user_id)
+
+
```
mfogel, line 139: I think maybe some logic from `master` got mixed in to your PR? I'm guessing including this was unintentional.

real-main/app/models/like/dynamo.py
```diff
@@ -27,6 +27,18 @@ def parse_pk(self, pk):
     def get_like(self, liked_by_user_id, post_id):
         return self.client.get_item(self.pk(liked_by_user_id, post_id))
 
+    def get_common_likes(self, liked_by_user_id):
+        query_kwargs = {
+            'KeyConditionExpression': Key('gsiA1PartitionKey').eq(f'like/{liked_by_user_id}'),
+            'IndexName': 'GSI-A1',
+        }
+        liked_users = self.client.generate_all_query(query_kwargs)
```
mfogel, line 35: I like where I think you're going with this (match users by things they've liked of other users). However, we don't want to return _all_ the matches to the client as this is an unbounded list. We would want to return a paginated list. For example, here's the graphql response for another such unbounded list: https://github.com/real-social-media/backend/blob/v2020-08-20/real-main/schema.graphql#L798-L801

real-main/app/models/user/dynamo/base.py
```diff
@@ -61,6 +61,10 @@ def add_user(
             query_kwargs['Item']['email'] = email
         if phone:
             query_kwargs['Item']['phoneNumber'] = phone
+        if date_of_birth:
+            query_kwargs['Item']['date_of_birth'] = date_of_birth
```
mfogel, line 65: We're using camelCase for dynamo attributes, as such this would be better as `dateOfBirth` rather than `date_of_birth`

real-main/app/models/user/dynamo/base.py
```diff
@@ -244,6 +252,17 @@ def generate_user_ids_by_subscription_level(self, sub_level, max_expires_at=None
...
+    def update_last_post_view_at(self, user_id, now=None):
+        now = now or pendulum.now('utc')
+        query_kwargs = {
+            'Key': self.pk(user_id),
+            'UpdateExpression': 'SET lastPostViewAt = :lpva',
+            'ConditionExpression': 'NOT lastPostViewAt > :lpva',
+            'ExpressionAttributeValues': {':lpva': now.to_iso8601_string()},
+        }
+        failure_warning = f'Failed to update lastPostViewAt for user `{user_id}`'
+        return self.client.update_item(query_kwargs, failure_warning=failure_warning)
```
mfogel, line 264: More unrelated changes... I assume this came over from `master`

real-main/schema.graphql
```diff
@@ -110,7 +110,9 @@ type Mutation {
     sharingDisabled: Boolean,
     verificationHidden: Boolean,
     languageCode: String,
-    themeCode: String
+    themeCode: String,
+    date_of_birth: AWSDate,
+    gender: String
```
mfogel, line 115: Nice job typing the `date_of_birth` with `AWSDate`!

I would also consider typing `gender` with a graphql enum with choices `M` and `F` and whatever other genders we wish to support.

Finally, we are using camelCase for our graphql schema - as such this would be better as `dateOfBirth` rather than `date_of_birth`.

real-main/app/handlers/appsync/handlers.py
```diff
@@ -24,6 +24,7 @@
 from app.models.user.enums import UserStatus
 from app.models.user.exceptions import UserException
 from app.utils import image_size
+from app.models import LikeManager, PotentialManager
```
mfogel, line 27: I didn't find `PotentialManager` defined anywhere in this PR

real-main/app/handlers/appsync/handlers.py
```diff
@@ -60,6 +63,7 @@
 comment_manager = managers.get('comment') or models.CommentManager(clients, managers=managers)
 follower_manager = managers.get('follower') or models.FollowerManager(clients, managers=managers)
 like_manager = managers.get('like') or models.LikeManager(clients, managers=managers)
+match_manager = managers.get('potential') or models.MatchManager(clients, managers=managers)
```
mfogel, line 66: To follow the pattern of the lines around it, this would be better as

```
match_manager = managers.get('match') or models.MatchManager(clients, managers=managers)
```

real-main/serverless/mapping-templates/mutation.yml
```diff
@@ -82,6 +82,18 @@
...
+  field: potential_matches
+  dataSource: LambdaDataSource
+  request: Lambda.request.vtl
+  response: Lambda.response.vt
+
+- type: Mutation
+  field: like_matches
+  dataSource: LambdaDataSource
+  request: Lambda.request.vtl
+  response: Lambda.response.vt
```
mfogel, line 95: I don't think these are mutations, I think these would both fit pretty nicely as fields on the user object. For example: `User.likeMatches` and `User.potentialMatches` could both be paginated lists of matches.

heroclass728:
Thanks for your comment and getting to know that you are a great dev person.
I needed to analyze master logic to move further, hence a bit confused.
But I am not disappointed because I have already made a lot of progress.
Regards,

PR #54, opened 2020-08-24
Add new two user fields and mapping templates

zkhin:


real-main/mapping-templates/Query.potentialMatches/after.response.vtl
```diff
@@ -0,0 +1,16 @@
...
+#end
+#set ($user = $ctx.prev.result[0])
+#set ($keys = [])
+
+$util.qr($keys.add({
+  'partitionKey': {'S': "userDateOfBirth/$user.dateOfBrith"},
+  'sortKey': {'S': '-'}
+}))
+
+$util.toJson($userIds)
```
mfogel, line 16: Nice job writing some VTL!

real-main/mapping-templates/User.dateOfBirth.request.vtl
```diff
@@ -0,0 +1,9 @@
+## email is private to the user themselves
```
mfogel, line 1: comment needs updating - no `email` in this template

real-main/schema.graphql
```diff
@@ -110,7 +113,9 @@ type Mutation {
     sharingDisabled: Boolean,
     verificationHidden: Boolean,
     languageCode: String,
-    themeCode: String
+    themeCode: String,
+    dateOfBirth: AWSDateTime,
+    gender: String
```
mfogel, line 118: Nice job using typing `dateOfBirth` as something other than a string. I think even better than `AWSDateTime` would probably be an `AWSDate`.

As far as typing the `gender`, I would probably make a custom graphql enum with types `M` and `F` and whatever other genders we want to support.

mfogel:
Nice job on the PR, this is totally useable but closing in favor of https://github.com/real-social-media/backend/pull/51 as that one is closer to merge already (has tests and one round of feedback complete)

PR #57, opened 2020-09-02
Add post focused view feature

zkhin:
https://trello.com/c/30led4yc/3-post-focused-view

### Description

Implement a ‘focused viewed post’ mutation
If possible, this should be a flag so that the front-end can use the existing mutation. This may be difficult since the font-end reports multiple views per request.

This should increase the weight (or view count) of a post to help it trend better. A user should only be able to view a post once (visible thumbnail triggers a view) and have one additional “focused view” if they open that post fullscreen. (2 views per user).

Create a user property (LastFocusView) which has the DateTime of the last time the user client submitted a focused view (fullscreen post view).


integration-testing/tests/trending-new.sequential.test.js
```diff
@@ -600,3 +600,58 @@ test('Only first view of a post counts for trending', async () => {
...
+  await otherClient.mutate({
+    mutation: mutations.reportPostViews,
+    variables: {postIds: [postId1], viewType: 'FOCUS'},
+  })
+  await misc.sleep(2000)
+  await ourClient.query({query: queries.trendingPosts}).then(({data: {trendingPosts}}) => {
+    expect(trendingPosts.items).toHaveLength(2)
+    expect(trendingPosts.items[0].postId).toBe(postId1)
+    expect(trendingPosts.items[1].postId).toBe(postId2)
+  })
```
mfogel, line 656: This is good, but I'm not sure it's sensitive to the last view reported being a `FOCUS` view. Is it? Like if you remove `viewType: 'FOCUS'` from line 649, does the test fail? If it does - please ignore the rest of this comment.

What I'd like to test here is that `multiplier` that FOCUS views get for trending. I think this would work to do so:
1. theirClient -> records a normal view on post2
1. otherClient -> records a normal view on post2
1. ourClient -> checks posts in trending, should just see post2
1. theirClient -> records a view on post1 with `viewType: FOCUS`
1. ourClient -> check posts in trending, should see post1 first then post2

I think that order of operations is sensitive to the last view reported being `FOCUS` and not `THUMBNAIL` / normal.

integration-testing/tests/posts/views.test.js
```diff
@@ -358,3 +358,74 @@ test('Post views and deleted on user delete/reset', async () => {
...
+
+  // verify the two posts have the right viewedBy lists
+  resp = await ourClient.query({query: queries.post, variables: {postId: postId1}})
+  expect(resp.data.post.viewedBy.items).toHaveLength(1)
+  expect(resp.data.post.viewedBy.items[0].userId).toBe(other1UserId)
+  resp = await ourClient.query({query: queries.post, variables: {postId: postId2}})
+  expect(resp.data.post.viewedBy.items).toHaveLength(2)
+  expect(resp.data.post.viewedBy.items[0].userId).toBe(other1UserId)
+  expect(resp.data.post.viewedBy.items[1].userId).toBe(other2UserId)
+})
```
mfogel, line 431: In this file - do you mind updating the syntax of these tests to match that in the other file? What I mean is:

Change stanzas like this:

```js
resp = await ourClient.query({query: queries.post, variables: {postId: postId1}})
expect(resp.data.post.viewedBy.items).toHaveLength(1)
expect(resp.data.post.viewedBy.items[0].userId).toBe(other1UserId)
```

to this

```js
await ourClient.query({query: queries.post, variables: {postId: postId1}}).then(({data: {post}}) => {
  expect(post.viewedBy.items).toHaveLength(1)
  expect(post.viewedBy.items[0].userId).toBe(other1UserId)
})
```

I've been trying to do my new tests in the 2nd format, I find it clearer and there's less function-wide state to keep track of.

real-main/app/mixins/view/model.py
```diff
@@ -32,18 +32,18 @@ def get_viewed_status(self, user_id):
...
-            self.view_dynamo.increment_view_count(self.id, user_id, view_count, viewed_at)
+            self.view_dynamo.increment_view_count(self.id, user_id, view_count, viewed_at, view_type)
         else:
             try:
-                self.view_dynamo.add_view(self.id, user_id, view_count, viewed_at)
+                self.view_dynamo.add_view(self.id, user_id, view_count, viewed_at, view_type)
             except ViewAlreadyExists:
                 # we lost a race condition to add the view, so still need to record our data
-                self.view_dynamo.increment_view_count(self.id, user_id, view_count, viewed_at)
+                self.view_dynamo.increment_view_count(self.id, user_id, view_count, viewed_at, view_type)
```
mfogel, line 46: I hate wasting characters, but do you mind changing the arguments passed in these three calls to `view_type=view_type`? Honestly I think it's a language design mistake that python allows you to pass keyword arguments using positions.

Some good bedtime reading for more backgorund on this, if you haven't already seen it: https://www.python.org/dev/peps/pep-0570/ https://www.python.org/dev/peps/pep-3102/

integration-testing/tests/trending-new.sequential.test.js
zkhin: @mfogel I modified the new test case following your steps but it's failing. Can you look into the issue if you have some time? Thank you
https://github.com/real-social-media/backend/blob/a9fcecd6abd199569a9e8be9e380663a5d3f1965/integration-testing/tests/trending-new.sequential.test.js#L604~L643

mfogel: I think there's an issue with line 629

```js
  await ourClient.query({query: queries.trendingPosts}).then(({data: {trendingPosts}}) => {
    expect(trendingPosts.items).toHaveLength(2)
    expect(trendingPosts.items[0].postId).toBe(postId2)
  })
```
There's only one post in the trending index at this point, so I think that should be `toHaveLength(1)`

mfogel:
Ok, so I had a chance to investigate in my local dev env.

I think we need to widen the scope of this PR a bit to make things work. One of the business requirements of this is that views of a given user on a given post should count for trending only for the first view of each type. So the first time a user looks at a thumbnail of a post, the post gets a 'trending point'. And the first time a user focus-views that same post, the post gets two more 'trending points'. But then on subsequent views of the post, either normal or FOCUS, the post should not get any more trending points.

I think this will be easier for us to implement using dynamo streams. So, suggested implementation:

- delete this stanza https://github.com/real-social-media/backend/blob/v2020-09-01/real-main/app/models/post/model.py#L616-L621
- add a method to the `PostManager`: `on_post_view_change_update_trending(post_id, new_item, old_item=None)`
    - this method should look at the new_item and old_item (which will be post view items) and figure out if the difference between the two items represents the first view of a type - either FOCUS or THUMBNAIL/no type.
    - if you grep through the code for `def on_` you will get dozens if not hundreds of examples of similar methods
    - if it is the first view of a type, then it should call `self.init_post(new_item)` to get a `post` object, then call `get_trending_multipler()`, and then increment the trending score.
- add a new `register()` call to `app/handlers/dynamo/handlers.py`
  - We want to register for the `INSERT` and `MODIFY` events: `register('post', 'view', ['INSERT', 'MODIFY'], post_manager.on_post_view_change_update_trending)`

Ok, let me know if you have questions, thank you - cheers

zkhin:
> And the first time a user focus-views that same post, the post gets two more 'trending points'. But then on subsequent views of the post, either normal or FOCUS, the post should not get any more trending points.

Thank you, I will figure out asap

real-main/app_tests/models/post/test_model.py
```diff
@@ -873,3 +874,6 @@ def test_get_trending_multiplier(post):
 
     post.user.item['subscriptionLevel'] = UserSubscriptionLevel.DIAMOND
     assert post.get_trending_multiplier() == 2
+
+    post.user.item['viewType'] = ViewType.FOCUS
+    assert post.get_trending_multiplier() == 2
```
mfogel, line 879: I think something got confused here. The user item doesn't have a field `viewType`. And `post.get_trending_multipler()` accepts a kwarg of `view_type`. This should probably be `assert post.get_trending_multipler(view_type=ViewType.FOCUS) == 4` ?

Could you also please add a test for when `view_type=ViewType.THUMBNAIL` ? Thank you

real-main/app_tests/models/post/test_model_record_view_count.py
```diff
@@ -47,7 +47,7 @@ def test_record_view_count_records_to_original_post_as_well(post, post2, user2):
     post.item['originalPostId'] = post2.id
     post.record_view_count(post.user_id, 1)
     assert post.view_dynamo.get_view(post.id, post.user_id)
-    assert post2.view_dynamo.get_view(post2.id, post.user_id) is None
+    assert post2.view_dynamo.get_view(post2.id, post.user_id)
```
mfogel, line 50: This whole stanza can be removed. We now treat post owner views the same as everyone else's views with respect to non-original posts.

real-main/app_tests/models/post/test_model_record_view_count.py
```diff
@@ -68,11 +68,11 @@ def test_text_only_posts_trend(post_manager, user, user2):
...
     post.record_view_count(user2.id, 4, viewed_at=viewed_at)
-    assert post.trending_score == 1 + 2
-    assert post.refresh_trending_item().trending_score == 1 + 2
+    assert post.trending_score == 1
+    assert post.refresh_trending_item().trending_score == 1
     assert user.refresh_trending_item()
-    assert pendulum.parse(user.trending_item['lastDeflatedAt']) == viewed_at
-    assert user.trending_score == 1
+    assert user.trending_item is None
+    assert user.trending_score is None
```
mfogel, line 75: This test case can be removed, it no longer tests anything.

real-main/app_tests/models/post/test_model_record_view_count.py
```diff
@@ -99,8 +99,8 @@ def test_only_first_view_of_post_causes_trending(post_manager, user, user2, user
     org_post_trending_score = post.trending_score
     org_user_trending_score = post.user.trending_score
     post.record_view_count(user3.id, 2)
-    assert post.refresh_trending_item().trending_score > org_post_trending_score
-    assert user.refresh_trending_item().trending_score > org_user_trending_score
+    assert post.refresh_trending_item().trending_score >= org_post_trending_score
+    assert user.refresh_trending_item().trending_score is None
```
mfogel, line 103: This test case can be removed

real-main/app_tests/models/post/test_model_record_view_count.py
```diff
@@ -124,9 +124,9 @@ def test_non_verified_image_posts_trend_with_lower_multiplier(post_manager, user
     # record a view, verify adds to trending
     viewed_at = pendulum.parse('2020-06-10T00:00:00Z')  # exactly one day forward
     post.record_view_count(user2.id, 4, viewed_at=viewed_at)
-    assert post.trending_score == 0.5 + 1
-    assert post.refresh_trending_item().trending_score == 0.5 + 1
-    assert user.refresh_trending_item().trending_score == 0.5  # includes an extra deflation compared to post
+    assert post.trending_score == 0.5
+    assert post.refresh_trending_item().trending_score == 0.5
+    assert user.refresh_trending_item().trending_score is None
```
mfogel, line 129: This test case can be remvoed

real-main/app_tests/models/post/test_model_record_view_count.py
```diff
@@ -143,9 +143,9 @@ def test_text_only_posts_trend_with_full_multiplier(post_manager, user, user2):
     # record a view, verify adds to trending
     viewed_at = pendulum.parse('2020-06-10T00:00:00Z')  # exactly one day forward
     post.record_view_count(user2.id, 4, viewed_at=viewed_at)
-    assert post.trending_score == 2 + 1
-    assert post.refresh_trending_item().trending_score == 2 + 1
-    assert user.refresh_trending_item().trending_score == 1  # includes an extra deflation compared to post
+    assert post.trending_score == 1
+    assert post.refresh_trending_item().trending_score == 1
+    assert user.refresh_trending_item().trending_score is None  # includes an extra deflation compared to post
```
mfogel, line 148: This test case can be removed

real-main/app_tests/models/post/test_model_record_view_count.py
```diff
@@ -163,9 +163,9 @@ def test_posts_from_subscriber_trend_with_boosted_multiplier(post_manager, user,
     # record a view, verify adds to trending
     viewed_at = pendulum.parse('2020-06-10T00:00:00Z')  # exactly one day forward
     post.record_view_count(user2.id, 4, viewed_at=viewed_at)
-    assert post.trending_score == 8 + 4
-    assert post.refresh_trending_item().trending_score == 8 + 4
-    assert user.refresh_trending_item().trending_score == 4  # includes an extra deflation compared to post
+    assert post.trending_score == 4
+    assert post.refresh_trending_item().trending_score == 4
+    assert user.refresh_trending_item().trending_score is None  # includes an extra deflation compared to post
```
mfogel, line 168: This test case can be removed

real-main/app_tests/models/post/test_model_record_view_count.py
```diff
@@ -214,5 +214,22 @@ def test_verified_image_posts_originality_determines_trending(post_manager, user
     assert user2.refresh_trending_item().trending_score is None
 
     # verify those trending points went to the original post & user
-    assert post.refresh_trending_item().trending_score == 1 + 2 + 2
-    assert user.refresh_trending_item().trending_score == 1 + 1
+    assert post.refresh_trending_item().trending_score == 1
+    assert user.refresh_trending_item().trending_score is None
```
mfogel, line 218: This test case can be removed

real-main/app_tests/models/post/test_model_record_view_count.py
```diff
@@ -214,5 +214,22 @@ def test_verified_image_posts_originality_determines_trending(post_manager, user
...
+    assert user.trending_score is None
+
+    # record a view, verify that boosts trending score
+    viewed_at = pendulum.parse('2020-06-10T00:00:00Z')  # exactly one day forward
+    post.record_view_count(user2.id, 4, viewed_at=viewed_at, view_type='FOCUS')
+    assert post.trending_score == 1
+    assert post.refresh_trending_item().trending_score == 1
+    assert user.refresh_trending_item()
+    assert user.trending_item is None
+    assert user.trending_score is None
```
mfogel, line 235: This test case can be removed

real-main/app_tests/models/user/test_dynamo.py
```diff
@@ -678,6 +678,14 @@ def test_update_last_post_view_at(user_dynamo, caplog):
     assert caplog.records[0].levelname == 'WARNING'
     assert all(x in caplog.records[0].msg for x in ['Failed to update lastPostViewAt', user_id_2])
 
+    # set it with view type
+    user_id_3 = str(uuid4())
+    user_dynamo.add_user(user_id_3, str(uuid4())[:8])
+    user_item = user_dynamo.update_last_post_view_at(user_id_3, None, 'FOCUS')
+    assert user_dynamo.get_user(user_id_3) == user_item
+    assert 'lastPostViewAt' in user_item
+    assert 'lastPostFocusViewAt' in user_item
```
mfogel, line 687: Please add a few things here:
 - before the call to `update_last_post_view_at`, please check to make sure `lastPostViewAt` and `lastPostFocusViewAt` are not in the user item
 - set a specific time in the call to `update_last_post_view_at()` (using the `now` kwarg)
 - after the call to `update_last_post_view_at()`, please verify the values of `lastPostViewAt` and `lastPostFocusViewAt` are correct (not just that the fields exist)

Finally, please add a short stanza to verify behavoir if `THUMBNAIL` is passed in for `view_type` (should set `lastPostViewAt`, but should not set `lastPostFocusViewAt`)

integration-testing/tests/posts/views.test.js
```diff
@@ -239,90 +279,105 @@ test('Post views on duplicate posts are viewed post and original post', async ()
...
-  expect(resp.data.post.postId).toBe(ourPostId)
-  expect(resp.data.post.viewedByCount).toBe(1)
-  expect(resp.data.post.viewedBy.items).toHaveLength(1)
-  expect(resp.data.post.viewedBy.items[0].userId).toBe(otherUserId)
+  await ourClient.query({query: queries.post, variables: {postId: ourPostId}}).then(({data: {post}}) => {
+    expect(post.postId).toBe(ourPostId)
+    expect(post.viewedByCount).toBe(1)
+    expect(post.viewedBy.items).toHaveLength(1)
+    expect(post.viewedBy.items[0].userId).toBe(otherUserId)
+  })
```
mfogel, line 380: Thanks for updating the format on all these test cases. 👍

mfogel (changes requested):

real-main/app_tests/models/post/test_manager_on.py
```diff
@@ -501,3 +501,22 @@ def test_on_post_view_add_delete_sync_viewed_by_counts(post_manager, post, caplo
...
+    trending_item = post_manager.trending_dynamo.get(post.id)
+    initial_score = float(trending_item.get('gsiA4SortKey'))
+    assert initial_score == pytest.approx(1 + 6 / 7, 0.1)
+    assert trending_item.get('gsiA4PartitionKey') == 'post/trending'
+
+    post_manager.on_post_view_change_update_trending(post.id, new_item=new_item)
+    trending_item = post_manager.trending_dynamo.get(post.id)
+    print(trending_item.get('gsiA4SortKey'))
+    assert float(trending_item.get('gsiA4SortKey')) == pytest.approx(initial_score * 3, 0.1)
+    assert trending_item.get('gsiA4PartitionKey') == 'post/trending'
```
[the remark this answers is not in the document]
zkhin:
```
called for first view of a post by a non-owner (either FOCUS or normal) and post.trending_increment_score returns False
```
Can you check why it doesn't return `False` in `post.trending_increment_score`?
https://github.com/real-social-media/backend/blob/91205b44e4ae86cb251e144fc17a1432a509d51d/real-main/app_tests/models/post/test_manager_on.py#L566~L571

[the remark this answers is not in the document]
mfogel: I pushed a commit to fix the situation with `post.trending_increment_score`, and flush out the unit tests a bunch in general.

What was going wrong with the version you had `assert post.trending_increment_score is False` is that this assumes that `trending_increment_score` is a property on the `post` object. It's not, it's a method with a return value.

I replaced that stanza with this test method: https://github.com/real-social-media/backend/pull/57/commits/07dde370d819333f82f4d179a044829a22a30430#diff-e6d8d5fa93a6a8526ea1119281372ab3R159

What this test method does is test the two lines here https://github.com/real-social-media/backend/pull/57/files#diff-d4e3d24ad23ff4defc2c46909a30cc5aR399-R401. by mocking the return value of `post.trending_increment_score` to make sure that the user's trending item is only updated if the posts is.

mfogel:
Ok - everything here looks good now. Sorry to add 'one more task' to this but before we merge it in, we should add a data migration to fill in `thumbnailViewCount` for existing post view items so that moving forward all of our post view items have consistent format. Without the data migration, we would have to remember that some post view items have correct `thumbnailViewCount` and others don't... better to make them all consistent.

So, I would create two new files:
- `real-main/migrations/post_view_0_4_add_thumbnail_view_count.py`
- `real-main/migrations_test/test_post_view_0_4_add_thumbnail_view_count.py`

I would use the following files as examples. I would copy-paste these into place and then just edit them from there
- `real-main/migrations/post_view_0_2_add_gsiA1_gsiA2.py`
- `real-main/migrations_test/test_post_view_0_2_add_gsiA1_gsiA2.py`

The new migration should:
- iterate over all post view items that do not have the `thumbnailViewCount` attribute
- for each of those items
  - calculate the correct `thumbnailViewCount` (`viewCount` - `focusViewCount`)
  - update the item (with appropriate `ConditionExpression` to make sure the counts haven't changed in the DB while the migration was working with the item in memory)

Note that this migration will not be run before the deployment - it will be run asynchronously after the deployment. The order of operations when I deploy this will be:
- update code and resources (`serverless deploy`)
- hang out for a few minutes, make sure the app hasn't broken (so new 'FOCUS' and 'THUMBNAIL' views will come in during this time)
- then run the migration, which will probably take about 10 minutes to run

real-main/migrations/post_view_0_4_add_thumbnail_view_count.py
```diff
@@ -0,0 +1,86 @@
...
+        thumbnail_view_count = view_count - focus_view_count
+        kwargs = {
+            'Key': {k: pv[k] for k in ('partitionKey', 'sortKey')},
+            'UpdateExpression': 'SET #vc = :vc, #fvc = :fvc, #tvc = :tvc',
+            'ConditionExpression': ' AND '.join(
+                [
+                    'attribute_exists(partitionKey)',
+                    'attribute_exists(viewCount)',
+                ]
+            ),
```
mfogel, line 57: In the `UpdateExpression`, we only want to set `thumbnailViewCount` - the other attributes should already be correct.

In the `ConditionExpression`, we should check the value of `viewCount` and `focusViewCount`. Example:

```py
            'ConditionExpression': ' AND '.join(
                [
                    'attribute_exists(partitionKey)',
                    '#vc = :vc',
                    '#fvc = :fvc',
                ]
            ),
```

The `attribute_exists(partitionKey)` in the `ConditionExpression` isn't strictly necessary since the other conditions will fail if the item no longer exists. It's fine to either leave it or remove it.

real-main/migrations/post_view_0_4_add_thumbnail_view_count.py
```diff
@@ -0,0 +1,86 @@
...
+            self.migrate_post_view(pv)
+
+    def generate_all_post_views(self):
+        "Return a generator of all items in the table that pass the filter"
+        scan_kwargs = {
+            'FilterExpression': ' AND '.join(
+                [
+                    'begins_with(partitionKey, :pk_prefix)',
+                    'begins_with(sortKey, :sk_prefix)',
+                    'attribute_not_exists(thumbnailViewCount)',
```
mfogel, line 30: This is my fault not yours - could you actually please remove this `attribute_not_exists(thumbnailViewCount)` here? I realized that we still need to iterate over items that have a `thumbnailViewCount` because some thumbnail views might come in from the frontend between the time of deployment and when this migration is run.

real-main/migrations/post_view_0_4_add_thumbnail_view_count.py
```diff
@@ -0,0 +1,86 @@
...
+            if 'LastEvaluatedKey' not in paginated:
+                break
+            scan_kwargs['ExclusiveStartKey'] = paginated['LastEvaluatedKey']
+
+    def migrate_post_view(self, pv):
+        post_id = pv['partitionKey'].split('/')[1]
+        user_id = pv['sortKey'].split('/')[1]
+        view_count = pv.get('viewCount', 0)
+        focus_view_count = pv.get('focusViewCount', 0)
+        thumbnail_view_count = view_count - focus_view_count
```
mfogel, line 48: Could we add here something like

```python
if pv.get('thumbnailViewCount', 0) == thumbnail_view_count:
  return
```

So that for items that don't need updating, we don't bother to do a write to the database.

mfogel (changes requested):
Ok, I added three comments to the migration - please address when you have time (along with any needed changes to the migration test.

Thank you very much!

PR #60, opened 2020-09-10
Card push notification diamond subscriber

zkhin:
Fixes: https://trello.com/c/BL6xiipb/7-new-card-push-notification-when-user-becomes-a-diamond-subscriber

### DIFF
--- a/integration-testing/tests/cards/user-subscription-level.test.js
+++ b/integration-testing/tests/cards/user-subscription-level.test.js
@@ -1,0 +1,41 @@
+/* eslint-env jest */
+
+// the aws-appsync-subscription-link pacakge expects WebSocket to be globaly defined, like in the browser
+global.WebSocket = require('ws')
+
+const cognito = require('../../utils/cognito')
+const {mutations, queries} = require('../../schema')
+
+const loginCache = new cognito.AppSyncLoginCache()
+
+beforeAll(async () => {
+  loginCache.addCleanLogin(await cognito.getAppSyncLogin())
+  loginCache.addCleanLogin(await cognito.getAppSyncLogin())
+})
+
+beforeEach(async () => await loginCache.clean())
+afterAll(async () => await loginCache.reset())
+
+test('Card message triggers cannot be called from external graphql client', async () => {
+  const {client: ourClient, userId: ourUserId} = await loginCache.getCleanLogin()
+
+  // we give ourselves some free diamond
+  await ourClient
+    .mutate({mutation: mutations.grantUserSubscriptionBonus})
+    .then(({data: {grantUserSubscriptionBonus: user}}) => {
+      expect(user.userId).toBe(ourUserId)
+      expect(user.subscriptionLevel).toBe('DIAMOND')
+    })
+
+  await ourClient.query({query: queries.self}).then(({data: {self: user}}) => {
+    expect(user.userId).toBe(ourUserId)
+    expect(user.cardCount).toBe(1)
+    expect(user.cards.items).toHaveLength(1)
+    let card = user.cards.items[0]
+    expect(card.cardId).toBeTruthy()
+    expect(card.cardId).toBe(`${ourUserId}:USER_SUBSCRIPTION_LEVEL`)
+    expect(card.title).toBe('Welcome to Diamond')
+    expect(card.subTitle).toBe('Enjoy exclusive perks of being a subscriber')
+    expect(card.action).toBe('https://real.app/diamond/')
+  })
+})
--- a/integration-testing/tests/cards/user-subscription-level.test.js
+++ b/integration-testing/tests/cards/user-subscription-level.test.js
@@ -1,22 +1,17 @@
-/* eslint-env jest */
-
-// the aws-appsync-subscription-link pacakge expects WebSocket to be globaly defined, like in the browser
-global.WebSocket = require('ws')
-
 const cognito = require('../../utils/cognito')
+const misc = require('../../utils/misc')
 const {mutations, queries} = require('../../schema')
 
 const loginCache = new cognito.AppSyncLoginCache()
 
 beforeAll(async () => {
   loginCache.addCleanLogin(await cognito.getAppSyncLogin())
-  loginCache.addCleanLogin(await cognito.getAppSyncLogin())
 })
 
 beforeEach(async () => await loginCache.clean())
 afterAll(async () => await loginCache.reset())
 
-test('Card message triggers cannot be called from external graphql client', async () => {
+test('User subscription level card: generating, format', async () => {
   const {client: ourClient, userId: ourUserId} = await loginCache.getCleanLogin()
 
   // we give ourselves some free diamond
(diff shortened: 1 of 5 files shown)


integration-testing/tests/cards/user-subscription-level.test.js
```diff
@@ -0,0 +1,41 @@
...
+
+beforeAll(async () => {
+  loginCache.addCleanLogin(await cognito.getAppSyncLogin())
+  loginCache.addCleanLogin(await cognito.getAppSyncLogin())
+})
+
+beforeEach(async () => await loginCache.clean())
+afterAll(async () => await loginCache.reset())
+
+test('Card message triggers cannot be called from external graphql client', async () => {
```
mfogel, line 19:
```suggestion
test('User subscription level card: generating, format', async () => {
```

integration-testing/tests/cards/user-subscription-level.test.js
```diff
@@ -0,0 +1,41 @@
...
+global.WebSocket = require('ws')
+
+const cognito = require('../../utils/cognito')
+const {mutations, queries} = require('../../schema')
+
+const loginCache = new cognito.AppSyncLoginCache()
+
+beforeAll(async () => {
+  loginCache.addCleanLogin(await cognito.getAppSyncLogin())
+  loginCache.addCleanLogin(await cognito.getAppSyncLogin())
```
mfogel, line 13: Since the most number of users being used in any of the tests in this file is just one, the `loginCache` only needs one entry. So only one call to `addCleanLogin()` is needed.

integration-testing/tests/cards/user-subscription-level.test.js
```diff
@@ -0,0 +1,41 @@
+/* eslint-env jest */
+
+// the aws-appsync-subscription-link pacakge expects WebSocket to be globaly defined, like in the browser
+global.WebSocket = require('ws')
```
mfogel, line 4: This is only needed for test suites that use graphql subscriptions. It can be removed from this test suite.

integration-testing/tests/cards/user-subscription-level.test.js
```diff
@@ -0,0 +1,41 @@
...
+      expect(user.userId).toBe(ourUserId)
+      expect(user.subscriptionLevel).toBe('DIAMOND')
+    })
+
+  await ourClient.query({query: queries.self}).then(({data: {self: user}}) => {
+    expect(user.userId).toBe(ourUserId)
+    expect(user.cardCount).toBe(1)
+    expect(user.cards.items).toHaveLength(1)
+    let card = user.cards.items[0]
+    expect(card.cardId).toBeTruthy()
```
mfogel, line 35: This `expect(card.cardId).toBeTruthy()` probably isn't necessary because the next line verifies it's a specific string.

real-main/app/models/card/templates.py
```diff
@@ -132,7 +132,7 @@ def __init__(self, user_id, post_id):
 
 class RequestedFollowersCardTemplate(CardTemplate):
 
-    action = 'https://real.app/chat/'
+    action = f'https://real.app/chat/'
```
mfogel: No need for this to be an f-string. This is probably why the python lint github action is failing on this PR.

real-main/app/models/card/templates.py
```diff
@@ -144,3 +144,19 @@ def __init__(self, user_id, requested_followers_count):
         self.card_id = self.get_card_id(user_id)
         cnt = requested_followers_count
         self.title = f'You have {cnt} pending follow request{"s" if cnt > 1 else ""}'
+
+
+class UserSubscriptionLevelTemplate(CardTemplate):
+    
+    action = f'https://real.app/diamond/'
```
mfogel:
```suggestion
    action = 'https://real.app/diamond'
```

real-main/app/models/card/templates.py
```diff
@@ -144,3 +144,19 @@ def __init__(self, user_id, requested_followers_count):
         self.card_id = self.get_card_id(user_id)
         cnt = requested_followers_count
         self.title = f'You have {cnt} pending follow request{"s" if cnt > 1 else ""}'
+
+
+class UserSubscriptionLevelTemplate(CardTemplate):
+    
+    action = 'https://real.app/diamond/'
```
mfogel, line 151:
```suggestion
    action = 'https://real.app/diamond'
```

This is for consistency with (most) of the other card actions.

integration-testing/tests/cards/user-subscription-level.test.js
```diff
@@ -0,0 +1,41 @@
...
+
+  // we give ourselves some free diamond
+  await ourClient
+    .mutate({mutation: mutations.grantUserSubscriptionBonus})
+    .then(({data: {grantUserSubscriptionBonus: user}}) => {
+      expect(user.userId).toBe(ourUserId)
+      expect(user.subscriptionLevel).toBe('DIAMOND')
+    })
+
+  await ourClient.query({query: queries.self}).then(({data: {self: user}}) => {
```
mfogel, line 30: Before this query, I would add an `await misc.sleep(2000)`

The reason for this is because the card is created with via an asynchronous dynamo stream handler. So there will be a delay of X ms before the card shows up. Usually X is fast - like say 100ms or less. But enough of the time it takes longer that if we don't put in these sleep calls it becomes impossible to run the whole integration test suite.

real-main/app_tests/models/card/test_manager_on.py
```diff
@@ -443,3 +444,22 @@ def test_on_post_viewed_by_count_change_update_card(card_manager, post, user):
...
+    user.item['subscriptionLevel'] = UserSubscriptionLevel.DIAMOND
+    card_manager.on_user_subscription_level_change_update_card(user.id, new_item=user.item)
+    assert card_manager.get_card(template.card_id)
+
+    # change from diamond to basic level, process, check card deleted
+    old_item = user.item.copy()
+    user.item['subscriptionLevel'] = UserSubscriptionLevel.BASIC
+    old_item['subscriptionLevel'] = UserSubscriptionLevel.DIAMOND
+    card_manager.on_user_subscription_level_change_update_card(user.id, new_item=user.item, old_item=old_item)
+    assert card_manager.get_card(template.card_id) is None
```
mfogel, line 465: This is good and all correct, but the flow is a bit different that what our users actually go through in our app. Could we do the following:

- verify card does not exist (you already have this)
- fires for user created in `BASIC`, verify card is _not_ created
- fires for user edited to get `DIAMOND`, verify card is created
- fires for user edited to go back to `BASIC`, verify card is deleted

Thank you!

mfogel (changes requested):

real-main/app_tests/models/card/test_manager_on.py
```diff
@@ -443,3 +444,27 @@ def test_on_post_viewed_by_count_change_update_card(card_manager, post, user):
...
+
+    # change to basic level, process, check card is not created
+    user.item['subscriptionLevel'] = UserSubscriptionLevel.BASIC
+    card_manager.on_user_subscription_level_change_update_card(user.id, new_item=user.item)
+    assert card_manager.get_card(template.card_id) is None
+
+    # change to diamond level, process, check card created
+    user.item['subscriptionLevel'] = UserSubscriptionLevel.DIAMOND
+    card_manager.on_user_subscription_level_change_update_card(user.id, new_item=user.item)
```
mfogel, line 462: This simulates a call as if a new user is _created_ with `DIAMOND`. It's ok to test that but that, and good that the handler here handles it ok, but that flow doesn't actually happen in our app, at least not right now.

What is important to test is when a user _changes_ from `BASIC` to `DIAMOND`.

Recommendation, keeping with the style of the rest of this test:

```py
old_item = user.item.copy()
user.item['subscriptionLevel'] = UserSubscriptionLevel.DIAMOND
card_manager.on_user_subscription_level_change_update_card(user.id, new_item=user.item, old_item=old_item)
```

real-main/app_tests/models/card/test_manager_on.py
```diff
@@ -443,3 +444,27 @@ def test_on_post_viewed_by_count_change_update_card(card_manager, post, user):
...
+    assert card_manager.get_card(template.card_id)
+
+    # change from diamond to basic level, process, check card deleted
+    old_item = user.item.copy()
+    user.item['subscriptionLevel'] = UserSubscriptionLevel.BASIC
+    old_item['subscriptionLevel'] = UserSubscriptionLevel.DIAMOND
```
mfogel, line 468: This line isn't needed... `old_item['subscriptionLevel']` is already `DIAMOND`. If you want to assert that here that has some use, but no need to set it again.

mfogel (changes requested):

PR #62, opened 2020-09-12
New Card for new full user without profile photo

zkhin:
Fixes: https://trello.com/c/5gOhvGcp/12-new-card-for-new-full-users-without-a-profile-photo

### DIFF
--- a/real-main/app/models/card/manager.py
+++ b/real-main/app/models/card/manager.py
@@ -4,7 +4,7 @@ from functools import partialmethod
 import pendulum
 
 from app import models
-from app.models.user.enums import UserSubscriptionLevel
+from app.models.user.enums import UserStatus, UserSubscriptionLevel
 
 from . import templates
 from .appsync import CardAppSync
@@ -145,6 +145,23 @@ class CardManager:
             new_subscription_level != UserSubscriptionLevel.DIAMOND
             and old_subscription_level == UserSubscriptionLevel.DIAMOND
         ):
+            self.dynamo.delete_card(card_template.card_id)
+
+    def on_user_change_update_photo_card(self, user_id, new_item, old_item=None):
+        new_photo_post_id = new_item.get('photoPostId')
+        old_photo_post_id = (old_item or {}).get('photoPostId')
+
+        new_user_status = new_item.get('userStatus')
+        old_user_status = (old_item or {}).get('userStatus')
+
+        card_template = templates.AddProfilePhotoCardTemplate(user_id)
+        if (
+            new_user_status == UserStatus.ACTIVE
+            and not new_photo_post_id
+            and (not old_item or (old_item and old_user_status == UserStatus.ANONYMOUS))
+        ):
+            self.add_or_update_card(card_template)
+        elif new_photo_post_id and not old_photo_post_id:
             self.dynamo.delete_card(card_template.card_id)
 
     on_user_followers_requested_count_change_sync_card = partialmethod(
--- a/real-main/app/models/card/manager.py
+++ b/real-main/app/models/card/manager.py
@@ -148,20 +148,20 @@ class CardManager:
             self.dynamo.delete_card(card_template.card_id)
 
     def on_user_change_update_photo_card(self, user_id, new_item, old_item=None):
-        new_photo_post_id = new_item.get('photoPostId')
-        old_photo_post_id = (old_item or {}).get('photoPostId')
+        new_has_photo = 'photoPostId' in new_item
+        old_has_photo = 'photoPostId' in (old_item or {})
 
-        new_user_status = new_item.get('userStatus')
-        old_user_status = (old_item or {}).get('userStatus')
+        new_user_status = new_item.get('userStatus', UserStatus.ACTIVE)
+        old_user_status = old_item.get('userStatus', UserStatus.ACTIVE) if old_item else None
 
         card_template = templates.AddProfilePhotoCardTemplate(user_id)
         if (
-            new_user_status == UserStatus.ACTIVE
-            and not new_photo_post_id
-            and (not old_item or (old_item and old_user_status == UserStatus.ANONYMOUS))
+            not new_has_photo
+            and new_user_status == UserStatus.ACTIVE
+            and old_user_status in (None, UserStatus.ANONYMOUS)
         ):
             self.add_or_update_card(card_template)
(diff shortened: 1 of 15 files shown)


real-main/app/models/card/manager.py
```diff
@@ -147,6 +147,23 @@ def on_user_subscription_level_change_update_card(self, user_id, new_item, old_i
         ):
             self.dynamo.delete_card(card_template.card_id)
 
+    def on_user_change_update_photo_card(self, user_id, new_item, old_item=None):
+        new_photo_post_id = new_item.get('photoPostId')
+        old_photo_post_id = (old_item or {}).get('photoPostId')
```
mfogel, line 152: Since we don't care about the value of the photo post id, and it would be a bummer to have a bug because of a `photoPostId` of `0` or `false`, I think it might make more sense to just keep track of a boolean rather than the actual post id value

```suggestion
        new_has_photo = 'photoPostId' in new_item
        old_has_photo = 'photoPostId' in (old_item or {})
```

real-main/app/models/card/manager.py
```diff
@@ -147,6 +147,23 @@ def on_user_subscription_level_change_update_card(self, user_id, new_item, old_i
...
+
+        card_template = templates.AddProfilePhotoCardTemplate(user_id)
+        if (
+            new_user_status == UserStatus.ACTIVE
+            and not new_photo_post_id
+            and (not old_item or (old_item and old_user_status == UserStatus.ANONYMOUS))
+        ):
+            self.add_or_update_card(card_template)
+        elif new_photo_post_id and not old_photo_post_id:
+            self.dynamo.delete_card(card_template.card_id)
```
mfogel, line 165: This can now be simplified because of the above changes to the local variables:

```suggestion
        if (
            not new_has_photo
            and new_user_status == UserStatus.ACTIVE
            and old_user_status in (None, UserStatus.ANONYMOUS)
        ):
            self.add_or_update_card(card_template)
        
        if new_has_photo and not old_has_photo:
            self.dynamo.delete_card(card_template.card_id)
```

I'm not a fan of `else if` constructs, I try to avoid them when possible. To know if an `else if` is true, you have to go back up the chain of `else if` and `if` statements and read the conditions in each of them. Better code locality using just plain `if` or `switch/case` in languages that support it (python doesn't 👎 )

real-main/app_tests/models/card/test_manager_on.py
```diff
@@ -469,3 +469,28 @@ def test_on_user_subscription_level_change_update_card(card_manager, user):
     user.item['subscriptionLevel'] = UserSubscriptionLevel.BASIC
     card_manager.on_user_subscription_level_change_update_card(user.id, new_item=user.item, old_item=old_item)
     assert card_manager.get_card(template.card_id) is None
+
+
+def test_on_user_change_update_photo_card(card_manager, user):
```
mfogel, line 474: Can you please test a few more scenarios (you can add them to this test or create a new one, either way is fine):

- creation of an `ANONYMOUS` user should not generate a card
- modifying a user that does not have a `photoPostId` from `ANONYMOUS` to `ACTIVE` _should_ generate a card

real-main/app_tests/models/card/test_manager_on.py
```diff
@@ -469,3 +469,28 @@ def test_on_user_subscription_level_change_update_card(card_manager, user):
...
+    user.item['photoPostId'] = str(uuid4())
+    card_manager.on_user_change_update_photo_card(user.id, new_item=user.item)
+    assert card_manager.get_card(template.card_id) is None
+
+    # add profile photo, process, check card deleted
+    old_item = user.item.copy()
+    user.item['photoPostId'] = str(uuid4())
+    user.item['userStatus'] = UserStatus.ACTIVE
+    card_manager.on_user_change_update_photo_card(user.id, new_item=user.item, old_item=old_item)
+    assert card_manager.get_card(template.card_id) is None
```
mfogel, line 496: The last stanza above this one ended with the following assert

```py
assert card_manager.get_card(template.card_id) is None
```

So the card doesn't exist when this stanza started, so we can't be sure this is actually deleting the card, right? It never existed in the first place.

You can fix this by either making separate test functions or by rearranging the stanzas, maybe adding more setup/configuration between them.

In general, the 'right' way to fix this would be to break this one test function up into like 5 separate smaller ones. I've been lazy in this project and written a lot of big unit test functions. But in general, unit tests are supposed to be short and minimal - as you no doubt already know. That being said, I don't really care if we do longer test functions in this project as long as I can fit each function on my screen all at once, which is about 50 lines 😊

mfogel (changes requested):

mfogel:
Oh - and I forgot to mention in the review. It looks like the python-lint job is failing... please try to fix that when you get a chance.

mfogel (changes requested):
Code is looking good! I left one minor comment in the last unit test, that's it.

But unfortunately, this change (by design) breaks a lot of the existing integration tests. When I run the full test suite I'm getting errors in the following files:

- tests/cards/user-subscription-level.test.js
- tests/cards/requested-followers.test.js
- tests/cards/post-likes.test.js
- tests/cards/chat.test.js
- tests/cards/comments.test.js
- tests/cards/post-views.test.js
- tests/cards/post-mention.test.js
- tests/cards/crud.test.js
- tests/cards/comment-mention.test.js
- tests/cards/post-repost.test.js

Nearly everything in the `integration-testing/test/cards` directory is failing.

What appears to be happening is that these failing tests are relying on the fact that new users don't have any cards. This PR changes that - now new ACTIVE users will have this card. So in general, when the failing tests do this:

```js
expect(user.cardCount).toBe(1)
expect(user.cards.items).toHaveLength(1)
expect(user.cards.items[0]).toBe(...)
```

They should instead do something like this:

```js
expect(user.cardCount).toBe(2)
expect(user.cards.items).toHaveLength(2)
// user.cards.items[0] is the 'Please add profile photo' card
expect(user.cards.items[1]).toBe(...)
```

Note: when integration tests with GQL subscriptions fail, jest can return several hundred (thousand?) lines of traceback, which isn't particularly helpful. You can cut that in half by using the `--noStackTrace` option

Also, in every file there is a `jest.retryTimes(2)` near the top - if you comment that out temporarily while running the tests the will give up after the first failure, rather than retry two more times.

Before we can merge this PR (or any other) all the integration tests need to pass.

PR #67, opened 2020-09-18
New user fields location match genders etc

zkhin:
Fixes: https://trello.com/c/NSNZv4VS/14-new-user-fields-currentlocation-matchgenders-matchagerange-matchlocationradius

### DIFF
--- a/integration-testing/schema/queries.js
+++ b/integration-testing/schema/queries.js
@@ -102,17 +102,6 @@ module.exports.self = gql`
       directChat {
         ...ChatFragment
       }
-      currentLocation {
-        latitude
-        longitude
-        accuracy
-      }
-      matchAgeRange {
-        min
-        max
-      }
-      matchGenders
-      matchLocationRadius
     }
   }
   ${fragments.album}
--- a/integration-testing/schema/queries.js
+++ b/integration-testing/schema/queries.js
@@ -102,6 +102,17 @@ module.exports.self = gql`
       directChat {
         ...ChatFragment
       }
+      currentLocation {
+        latitude
+        longitude
+        accuracy
+      }
+      matchAgeRange {
+        min
+        max
+      }
+      matchGenders
+      matchLocationRadius
     }
   }
   ${fragments.album}
--- a/integration-testing/tests/users/properties.test.js
+++ b/integration-testing/tests/users/properties.test.js
@@ -622,12 +622,6 @@ test('Set and read properties(currentLocation, matchAgeRange, matchGenders, matc
     },
   })
 
-  await ourClient.query({query: queries.self}).then(({data: {self: user}}) => {
-    expect(user.userId).toBe(userId)
-    expect(JSON.stringify(user.matchGenders)).toBe(JSON.stringify(matchGenders))
-    expect(user.matchLocationRadius).toBe(matchLocationRadius)
-  })
-
   // Set user age range
   await ourClient.mutate({
     mutation: mutations.setUserAgeRange,
@@ -635,12 +629,6 @@ test('Set and read properties(currentLocation, matchAgeRange, matchGenders, matc
       min: matchAgeRange.min,
       max: matchAgeRange.max,
     },
-  })
-
-  await ourClient.query({query: queries.self}).then(({data: {self: user}}) => {
-    expect(user.userId).toBe(userId)
-    expect(user.matchAgeRange.min).toBe(matchAgeRange.min)
-    expect(user.matchAgeRange.max).toBe(matchAgeRange.max)
   })
 
   // Set user current location
@@ -655,6 +643,10 @@ test('Set and read properties(currentLocation, matchAgeRange, matchGenders, matc
 
   await ourClient.query({query: queries.self}).then(({data: {self: user}}) => {
     expect(user.userId).toBe(userId)
+    expect(JSON.stringify(user.matchGenders)).toBe(JSON.stringify(matchGenders))
+    expect(user.matchLocationRadius).toBe(matchLocationRadius)
+    expect(user.matchAgeRange.min).toBe(matchAgeRange.min)
+    expect(user.matchAgeRange.max).toBe(matchAgeRange.max)
     expect(user.currentLocation.latitude).toBe(currentLocation.latitude)
     expect(user.currentLocation.longitude).toBe(currentLocation.longitude)
     expect(user.currentLocation.accuracy).toBe(currentLocation.accuracy)
--- a/integration-testing/tests/users/properties.test.js
+++ b/integration-testing/tests/users/properties.test.js
@@ -660,6 +660,20 @@ test('Set and read properties(currentLocation, matchAgeRange, matchGenders, matc
     expect(user.currentLocation).toBeNull()
     expect(user.matchAgeRange).toBeNull()
   })
+
+  // update current location without accuracy, process and check if accuracy is null
+  await ourClient.mutate({
+    mutation: mutations.setUserCurrentLocation,
+    variables: {
+      latitude: currentLocation.latitude,
+      longitude: currentLocation.longitude,
+    },
+  })
+
+  await ourClient.query({query: queries.user, variables: {userId}}).then(({data: {user}}) => {
+    expect(user.userId).toBe(userId)
+    expect(user.currentLocation.accuracy).toBeNull()
+  })
 })
 
 test('Validate user properties(currentLocation, matchAgeRange, matchGenders, matchLocationRadius)', async () => {
--- a/integration-testing/tests/users/properties.test.js
+++ b/integration-testing/tests/users/properties.test.js
@@ -605,6 +605,7 @@ test('User setUserAPNSToken', async () => {
 
 test('Set and read properties(currentLocation, matchAgeRange, matchGenders, matchLocationRadius)', async () => {
   const {client: ourClient, userId} = await loginCache.getCleanLogin()
+  const {client: theirClient} = await loginCache.getCleanLogin()
 
   // set up another user in cognito, leave them as public
   const currentLocation = {latitude: 50.01, longitude: 50.01, accuracy: 50}
@@ -657,6 +658,15 @@ test('Set and read properties(currentLocation, matchAgeRange, matchGenders, matc
     expect(user.currentLocation.latitude).toBe(currentLocation.latitude)
     expect(user.currentLocation.longitude).toBe(currentLocation.longitude)
     expect(user.currentLocation.accuracy).toBe(currentLocation.accuracy)
+  })
+
+  // check another user can't see values
+  await theirClient.query({query: queries.user, variables: {userId}}).then(({data: {user}}) => {
+    expect(user.userId).toBe(userId)
+    expect(user.matchGenders).toBeNull()
+    expect(user.matchLocationRadius).toBeNull()
+    expect(user.currentLocation).toBeNull()
+    expect(user.matchAgeRange).toBeNull()
   })
 })
 
--- a/integration-testing/tests/users/properties.test.js
+++ b/integration-testing/tests/users/properties.test.js
@@ -24,581 +24,608 @@ afterEach(async () => {
   anonClient = null
 })
 
-describe('Read and write properties our our own profile', () => {
-  // username is tested in the set-username.test.js
+// describe('Read and write properties our our own profile', () => {
+//   // username is tested in the set-username.test.js
 
-  test('followed/follwer status', async () => {
-    const {client, userId} = await loginCache.getCleanLogin()
-    let resp = await client.query({query: queries.user, variables: {userId}})
-    expect(resp.data.user.followedStatus).toBe('SELF')
-    expect(resp.data.user.followerStatus).toBe('SELF')
-  })
+//   test('followed/follwer status', async () => {
+//     const {client, userId} = await loginCache.getCleanLogin()
+//     let resp = await client.query({query: queries.user, variables: {userId}})
+//     expect(resp.data.user.followedStatus).toBe('SELF')
+//     expect(resp.data.user.followerStatus).toBe('SELF')
+//   })
 
-  test('privacyStatus', async () => {
-    const {client, userId} = await loginCache.getCleanLogin()
-    let resp = await client.query({query: queries.user, variables: {userId}})
-    expect(resp.data.user.privacyStatus).toBe('PUBLIC')
+//   test('privacyStatus', async () => {
+//     const {client, userId} = await loginCache.getCleanLogin()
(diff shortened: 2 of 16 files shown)


real-main/schema.graphql
```diff
@@ -124,7 +124,11 @@ type Mutation {
     languageCode: String,
     themeCode: String,
     birthday: AWSDate,
-    gender: UserGender
+    gender: UserGender,
+    currentLocation: LocationInput,
+    matchAgeRange: AgeRangeInput,
+    matchGenders: [UserGender!],
+    matchLocationRadius: Int
```
mfogel, line 131:
```suggestion
    matchLocationRadius: Int,
```

I think the lint rules are ok with trailing commas in situations like this... let's use one here if possible.

real-main/app/handlers/appsync/handlers.py
```diff
@@ -321,6 +330,33 @@ def set_user_details(caller_user, arguments, **kwargs):
...
+
+        if minAge > maxAge or minAge < 18 or maxAge > 100:
+            raise ClientException('Invalid matchAgeRange')
+
+    if match_location_radius is not None:
+        if match_location_radius < 15:
+            raise ClientException('MatchLocationRadius should be greater than or equal to 15')
+        if caller_user.subscription_level == UserSubscriptionLevel.BASIC and match_location_radius > 100:
+            raise ClientException('MatchLocationRadius should be less than or equal to 100')
+
```
mfogel, line 359: The code/logic here to do all this validation looks good, but the problem is that when it's in the handler function it is hard to unit test. So I think we should move it, and unit test it. (I realize there are integration tests for this, that's great, but we need unit tests also whenever possible 😊 )

Two options come to mind. I'm ok with either of these, though I think the second one is more the 'right' answer...

1. Move the logic to within `User.update_details()`. `User.updateDetails()` would raise a `UserException` (or a subclass of `UserException`, which would be caught in the handler function, which would then raise a `ClientException`. Unit tests would go in `app_tests/models/user/test_model.py`

1. Make a new file `app/handlers/appsync/validation.py` that has functions like `def validate_match_location_radius(value)`, `def validate_age_range(value)` etc. Those functions could throw a `ClientException` directly. They would of course be called from this file. You would also need a new file for unit tests for these functions: `app_tests/handlers/appsync/test_validation.py`

integration-testing/schema/queries.js
```diff
@@ -102,6 +102,17 @@ module.exports.self = gql`
...
+        latitude
+        longitude
+        accuracy
+      }
+      matchAgeRange {
+        min
+        max
+      }
+      matchGenders
+      matchLocationRadius
```
mfogel, line 115: Instead of adding this stanza here, could you add it to `integration-testing/schema/fragments.js` in the `simpleUserFields` fragment? Then it will also show up in the other queries, not just in `Query.self`

integration-testing/tests/users/properties.test.js
```diff
@@ -602,3 +602,122 @@ test('User setUserAPNSToken', async () => {
...
+    ourClient.mutate({
+      mutation: mutations.setUserCurrentLocation,
+      variables: {
+        longitude: 50.01,
+        latitude: 50.01,
+        accuracy: currentLocation.accuracy,
+      },
+    }),
+  ).rejects.toThrow(/ClientError: Accuracy should be greater than or equal to zero/)
+})
```
mfogel, line 723: We need integration test(s) to verify that when a user tries to read these fields on a different user, they just get `null` for each field. These fields should be private: I can only read my own values, I can't read anybody else's.

To get that to work, I think you're going to need to add resolvers and mapping templates for each of these fields using `NoneDataSource`. `User.email` looks like a good example to follow:
- https://github.com/real-social-media/backend/blob/v2020-09-15/real-main/mapping-templates/User.email.request.vtl
- https://github.com/real-social-media/backend/blob/v2020-09-15/real-main/serverless/mapping-templates/user.yml#L86-L89

mfogel (changes requested):
Nice work! Please address these comments when you have time.

Please also double-check to make sure all the existing integration tests pass... I suspect they don't? Specifically, I think this one probably fails? https://github.com/real-social-media/backend/blob/v2020-09-15/integration-testing/tests/block/reads.test.js#L23. I haven't tested it directly myself. There may be others that fail as well, not sure.

integration-testing/tests/users/properties.test.js
```diff
@@ -602,3 +602,132 @@ test('User setUserAPNSToken', async () => {
...
+  })
+
+  // check another user can't see values
+  await theirClient.query({query: queries.user, variables: {userId}}).then(({data: {user}}) => {
+    expect(user.userId).toBe(userId)
+    expect(user.matchGenders).toBeNull()
+    expect(user.matchLocationRadius).toBeNull()
+    expect(user.currentLocation).toBeNull()
+    expect(user.matchAgeRange).toBeNull()
+  })
```
mfogel, line 670: This is good, except it does not pass. The test is now correct, but the backend implementation is not.

To get this to work, I think you're going to need to add resolvers and mapping templates for each of these fields using NoneDataSource. User.email looks like a good example to follow:

- https://github.com/real-social-media/backend/blob/v2020-09-15/real-main/mapping-templates/User.email.request.vtl
- https://github.com/real-social-media/backend/blob/v2020-09-15/real-main/serverless/mapping-templates/user.yml#L86-L89

mfogel (changes requested):

integration-testing/tests/users/properties.test.js
zkhin: Oh, my bad. I committed files in the `integration-testing` directory so the new vtl files are not pushed yet...

real-main/app/handlers/appsync/handlers.py
```diff
@@ -321,6 +331,20 @@ def set_user_details(caller_user, arguments, **kwargs):
     if privacy_status is not None:
         caller_user.set_privacy_status(privacy_status)
 
+    if current_location is not None:
+        latitude = Decimal(str(current_location.get('latitude')))
+        longitude = Decimal(str(current_location.get('longitude')))
+        accuracy = Decimal(str(current_location.get('accuracy', 0)))
+
+        current_location = {"latitude": latitude, "longitude": longitude, "accuracy": accuracy}
```
mfogel, line 339: A few thoughts on this:

- Since using `Decimal` is essentially a dynamodb requirement, I think we should probably put this conversion 'closer' to dynamo. Probably in the method `set_user_details()` of `real-main/app/models/user/dynamo/base.py`
- if `accuracy` isn't set by the client, let's leave it not set in dynamo (don't fill in a zero)
- in the interest of 'fail early and often' I would do `current_location['latitude']`, not `current_location.get('latitude')`
- to convert a float to a decimal, I think this is a better formulation:

```py
from decimal import BasicContext
my_float = 6/7
my_decimal = BasicContext.create_decimal(my_float).normalize()
```

I'm going to go through other places in this project where we use `Decimal` and change them to initialize decimals in this way (will do that in a separate commit, not part of this PR)

real-main/mapping-templates/User.currentLocation.request.vtl
```diff
@@ -0,0 +1,9 @@
+## email is private to the user themselves
```
mfogel, line 1:
```suggestion
## private to the user themselves
```

This suggestion applies to the other mapping templates in this PR as well.

real-main/mapping-templates/User.currentLocation.request.vtl
```diff
@@ -0,0 +1,9 @@
...
+#if ($ctx.source.userId != $ctx.identity.cognitoIdentityId)
+  #return ()
```
mfogel, line 3:
```suggestion
  #return
```

Simpler way to write the same thing I discovered a few months ago. This suggestion applies to the other mapping templates in this PR as well.

mfogel (changes requested):

real-main/app/models/user/dynamo/base.py
```diff
@@ -168,6 +173,16 @@ def process_attr(name, value):
...
 
+        if current_location is not None:
+            latitude = BasicContext.create_decimal(current_location['latitude']).normalize()
+            longitude = BasicContext.create_decimal(current_location['longitude']).normalize()
+            accuracy = current_location.get('accuracy')
+
+            if accuracy is not None:
+                accuracy = BasicContext.create_decimal(accuracy).normalize()
+
+            current_location = {"latitude": latitude, "longitude": longitude, "accuracy": accuracy}
```
mfogel, line 184: This is better - but this will still sometimes try to write null/None into dynamo...

```suggestion
            current_location = {"latitude": latitude, "longitude": longitude}
            if 'accuracy' in current_location:
                current_location['accuracy'] = BasicContext.create_decimal(current_location['accuracy']).normalize()            
```

Please make sure the unit tests for this function check the situations where
- accuracy is not set
- accuracy is set

mfogel (commented):
I haven't reviewed the unit tests in detail yet, I will later today

real-main/app/models/user/dynamo/base.py
mfogel, line 184:
```suggestion
        for key in ('latitude', 'longitude'):
            current_location[key] = BasicContext.create_decimal(current_location[key]).normalize()
```

We actually don't need to apply the conversion to `Decimal` to accuracy because it is an integer, not a float.

mfogel: You may need to check and handle the situation where `accuracy` is in `current_location`, but has value `None`. I'm not sure if this is needed - it depends on how appsync/graphql handles the situation where `accuracy`:
  - is just missing from the GQL type
  - is in the GQL type with value `null`

So you might need to add something like this:

```py
if current_location.get('accuracy') is None:
  current_location.pop('accuracy', None)
```

real-main/app_tests/handlers/appsync/test_validation.py
```diff
@@ -0,0 +1,75 @@
...
+
+    # Pass the validation
+    assert validate_age_range(valid_match_age_range) is True
+
+    # Raise client exception
+    with pytest.raises(ClientException):
+        validate_age_range(invalid_match_age_range)
+
+
+def test_validate_current_location():
```
mfogel, line 46: Values I would make sure to test here:

- latitude:
  - -90
  - 90
  - 0
  - -90.1
  - 90.1
- longitude
  - -180
  - 180
  - 0
  - -180.1
  - 180.1
- accuracy
  - 0
  - some integer greater than 0... say 42
  - None
  - not present

real-main/app_tests/models/user/test_dynamo.py
```diff
@@ -275,10 +284,23 @@ def test_set_user_details(user_dynamo):
...
     assert resp == expected
 
+    # assert if accuracy is not set
+    current_location = {"latitude": Decimal('50'), "longitude": Decimal('50')}
+    resp = user_dynamo.set_user_details(
+        user_id,
+        current_location=current_location,
+    )
+    assert resp['currentLocation'] == current_location
+    assert resp['currentLocation'].get('accuracy') is None
```
mfogel, line 302:
```suggestion
    assert 'accuracy' not in resp['currentLocation']
```

This suggestion is a slightly stricter assertion - it should ensure we aren't storing a null in dynamo

real-main/app_tests/models/user/test_dynamo.py
```diff
@@ -232,6 +233,10 @@ def test_set_user_details(user_dynamo):
     username = 'my-username'
     birthday = pendulum.from_format('1900-01-01', 'YYYY-MM-DD').to_iso8601_string()
     gender = 'MALE'
+    current_location = {"latitude": Decimal('50'), "longitude": Decimal('50'), "accuracy": Decimal('50')}
+    match_age_range = {"min": Decimal('20'), "max": Decimal('50')}
+    match_genders = ['MALE', 'FEMALE']
+    match_location_radius = Decimal('15')
```
mfogel, line 239: I think all these should be floats or ints, not Decimals anymore

integration-testing/tests/users/properties.test.js
```diff
@@ -602,3 +602,138 @@ test('User setUserAPNSToken', async () => {
...
+
+  // Validate match location radius
+  await expect(
+    ourClient.mutate({
+      mutation: mutations.setUserDetails,
+      variables: {
+        matchLocationRadius: matchLocationRadius,
+      },
+    }),
+  ).rejects.toThrow(/ClientError: MatchLocationRadius should be greater than or equal to 15/)
```
mfogel, line 695:
```suggestion
  ).rejects.toThrow(/ClientError: matchLocationRadius should be greater than or equal to 15/)
```

integration-testing/tests/users/properties.test.js
```diff
@@ -602,3 +602,138 @@ test('User setUserAPNSToken', async () => {
...
+  // Validate user current location
+  await expect(
+    ourClient.mutate({
+      mutation: mutations.setUserCurrentLocation,
+      variables: {
+        latitude: currentLocation.latitude,
+        longitude: 50.01,
+      },
+    }),
+  ).rejects.toThrow('GraphQL error: ClientError: Latitude should be in [-90, 90]')
```
mfogel, line 717:
```suggestion
  ).rejects.toThrow(/ClientError: latitude should be in \[-90, 90\]/)
```

integration-testing/tests/users/properties.test.js
```diff
@@ -602,3 +602,138 @@ test('User setUserAPNSToken', async () => {
...
+
+  await expect(
+    ourClient.mutate({
+      mutation: mutations.setUserCurrentLocation,
+      variables: {
+        latitude: 50.01,
+        longitude: currentLocation.longitude,
+      },
+    }),
+  ).rejects.toThrow('GraphQL error: ClientError: Longitude should be in [-180, 180]')
```
mfogel, line 727:
```suggestion
  ).rejects.toThrow(/ClientError: longitude should be in \[-180, 180\]/)
```

integration-testing/tests/users/properties.test.js
```diff
@@ -602,3 +602,138 @@ test('User setUserAPNSToken', async () => {
...
+  await expect(
+    ourClient.mutate({
+      mutation: mutations.setUserCurrentLocation,
+      variables: {
+        longitude: 50.01,
+        latitude: 50.01,
+        accuracy: currentLocation.accuracy,
+      },
+    }),
+  ).rejects.toThrow(/ClientError: Accuracy should be greater than or equal to zero/)
```
mfogel, line 738:
```suggestion
  ).rejects.toThrow(/ClientError: accuracy should be greater than or equal to zero/)
```

mfogel (commented):
Minor fixes to get the integration tests to pass. I think after these changes it will be ready to merge in.

PR #69, opened 2020-09-24
Feature/enable disable dating for user

zkhin:
Fixes: https://trello.com/c/vSPnfgyb/16-enable-disable-dating-for-a-user


real-main/app/handlers/appsync/validation.py
```diff
@@ -32,3 +32,25 @@ def validate_current_location(current_location):
...
+        'matchLocationRadius',
+    ]
+    for field in required_fields:
+        value = user.item.get(field)
+
+        if field == 'userStatus' and value is UserStatus.ANONYMOUS:
+            raise ClientException('ANONYMOUS user cannot update dating status')
+        if field != 'userStatus' and value is None:
+            raise ClientException(f'`{field}` is required field')
+    return True
```
mfogel, line 56: This is good logic - but I think it should live on the user model, not here.

I would probably just put this in `User.set_dating_status()`. You could, if you want, create a new method on the user model something like `User.can_enable_dating()` and then call that method from `User.set_dating_status()`, but I probably wouldn't bother... I would probably just put all this logic in `User.set_dataing_status()`.

mfogel: Oh - and no need to validate on `userStatus` here. That's already handled by the decorator `@validate_caller` in `app/handlers/appsync/handlers.py`.  Anonymous users shouldn't be able to get this point.

real-main/app/handlers/appsync/handlers.py
```diff
@@ -401,6 +406,19 @@ def set_user_apns_token(caller_user, arguments, **kwargs):
     return caller_user.serialize(caller_user.id)
 
 
+@routes.register('Mutation.setUserDatingStatus')
+@validate_caller
+@update_last_client
+def set_user_dating_status(caller_user, arguments, **kwargs):
+    status = arguments.get('status')
```
mfogel, line 413:
```suggestion
    status = arguments['status']
```

fail early and often. `status` is required, so better if this fails on access than it pulls out `None` and then fails further down the code.

real-main/app/handlers/appsync/handlers.py
```diff
@@ -401,6 +406,19 @@ def set_user_apns_token(caller_user, arguments, **kwargs):
...
+
+    user = user_manager.get_user(caller_user.id)
+    validate_dating_status_access_permission(user)
+
+    caller_user.set_dating_status(status)
```
mfogel, line 418:
```suggestion
    try:
        caller_user.set_dating_status(status)
    except UserException as err:
        raise ClientException(str(err)) from err
```

`caller_user` has already been initalized from dynamo, no need to get it again.

And this validation logic, since it is dependent on the state of the user, should go within `set_dating_status()` (and raise a `UserException` if it fails).

mfogel (changes requested):

real-main/app/models/user/dynamo/base.py
```diff
@@ -412,3 +412,11 @@ def add_user_deleted(self, user_id, now=None):
...
         key = {'partitionKey': f'user/{user_id}', 'sortKey': 'deleted'}
         return self.client.delete_item(key)
+
+    def set_user_dating_status(self, user_id, status):
+        query_kwargs = {
+            'Key': self.pk(user_id),
+            'UpdateExpression': 'SET datingStatus = :ds',
+            'ExpressionAttributeValues': {':ds': status},
+        }
+        return self.client.update_item(query_kwargs)
```
mfogel, line 422: Since the default value of `datingStatus` is `DISABLED`, I think rather than setting `User.datingStatus` to `DISABLED` we should just remove the attribute.

By doing so we will only have user items in two states: `datingStatus = ENABLED` and `datingStatus` not set. Rather than three states, two of which are equivalent.

integration-testing/tests/users/set-dating-status.test.js
```diff
@@ -0,0 +1,144 @@
+const fs = require('fs')
+const path = require('path')
+const uuidv4 = require('uuid/v4')
+
+const cognito = require('../../utils/cognito')
+const {mutations, queries} = require('../../schema')
+
+let anonClient
```
mfogel, line 8: `anonClient` is not used in this file. No need to declare it here - this can be removed.

real-main/schema.graphql
```diff
@@ -402,6 +410,8 @@ type User {
   matchGenders: [UserGender!]
   matchLocationRadius: Int
 
+  datingStatus: DatingStatus
```
mfogel, line 413: Could you move this declaration up, to just under `subscriptionExpiresAt: AWSDateTime` ?

integration-testing/tests/users/set-dating-status.test.js
```diff
@@ -0,0 +1,144 @@
...
+
+  // Set fullName
+  await ourClient.mutate({mutation: mutations.setUserDetails, variables: {fullName: 'Hunter S'}})
+  await expect(
+    ourClient.mutate({
+      mutation: mutations.setUserDatingStatus,
+      variables: {
+        status: 'ENABLED',
+      },
+    }),
```
mfogel, line 46:
```suggestion
    ourClient.mutate({mutation: mutations.setUserDatingStatus, variables: {status: 'ENABLED'}}),
```

This applies to all calls to `mutations.setUserDatingStatus` in this file

integration-testing/tests/users/set-dating-status.test.js
```diff
@@ -0,0 +1,144 @@
...
+
+  // Set matchLocationRadius
+  await ourClient.mutate({
+    mutation: mutations.setUserDetails,
+    variables: {
+      matchLocationRadius: 20,
+    },
+  })
+  // All required fields are fulfilled
+  await ourClient.mutate({mutation: mutations.setUserDatingStatus, variables: {status: 'ENABLED'}})
```
mfogel, line 137:
```suggestion
  await ourClient
    .mutate({mutation: mutations.setUserDatingStatus, variables: {status: 'ENABLED'}})
    .then(({data: {setUserDatingStatus: user}}) => expect(user.datingStatus).toBe('ENABLED'))
```

(apologies if there are typos in this suggestion, I did not actually run it

integration-testing/tests/users/set-dating-status.test.js
```diff
@@ -0,0 +1,144 @@
...
+beforeEach(async () => await loginCache.clean())
+afterAll(async () => await loginCache.reset())
+afterEach(async () => {
+  if (anonClient) await anonClient.mutate({mutation: mutations.deleteUser})
+  anonClient = null
+})
+
+test('Validate user dating status permission', async () => {
+  const {client: ourClient, userId} = await loginCache.getCleanLogin()
+
```
mfogel, line 27: Should check here that by default new users have `datingStatus` set to `DISABLED`.

integration-testing/tests/users/set-dating-status.test.js
```diff
@@ -0,0 +1,144 @@
...
+  ).rejects.toThrow(/ClientError: `matchLocationRadius` is required field/)
+
+  // Set matchLocationRadius
+  await ourClient.mutate({
+    mutation: mutations.setUserDetails,
+    variables: {
+      matchLocationRadius: 20,
+    },
+  })
+  // All required fields are fulfilled
```
mfogel, line 136: Before doing the call that will actually change the dating status, it would be good to first verify it didn't already accidentially change status. IE - add something like

```js
  await ourClient.query({query: queries.user, variables: {userId}}).then(({data: {user}}) => {
    expect(user.userId).toBe(userId)
    expect(user.datingStatus).toBe('DISABLED')
  }
```

integration-testing/tests/users/set-dating-status.test.js
```diff
@@ -0,0 +1,144 @@
...
+  await ourClient.mutate({mutation: mutations.setUserDatingStatus, variables: {status: 'ENABLED'}})
+
+  // Check if the datingStatus field is ENABLED
+  await ourClient.query({query: queries.user, variables: {userId}}).then(({data: {user}}) => {
+    expect(user.userId).toBe(userId)
+    expect(user.datingStatus).toBe('ENABLED')
+  })
+})
```
mfogel, line 144: Other things we should test via integration tests:

- `datingStatus` can be set to `DISABLED`
- other users see our `datingStatus` as `null`

mfogel (changes requested):
Please address these comments when you have time, thanks.

Also - I suspect the test `Blocked user only see absolutely minimal profile of blocker via direct access` in `integration-testing/tests/block/reads.test.js` is failing. Can you check? I think you will need to add `datingStatus` to that test, similar to how other fields are handled.

mfogel:
@zkhin when you have time, could you please rebase this on master and then add a check to make sure the user has the dyanmo attribute `age` set? I merged this one https://trello.com/c/Yt510hy3/20-new-dynamo-user-profile-field-age into master just a few minutes ago.

Also, `age` must be `18 <= age <= 100` in order to enable dating. Thank you!

PR #70, opened 2020-09-28
Add new lastFoundUsers field

zkhin:
Fixes: https://trello.com/c/n4wCbKrl/2-new-user-property-to-show-if-user-has-found-friends-using-contact-list


integration-testing/tests/find-users.test.js
```diff
@@ -76,45 +77,33 @@ test('Find users by email', async () => {
...
+  // find them by just phone
+  await misc.sleep(2000)
+  await ourClient
+    .query({query: queries.findUsers, variables: {phoneNumbers: [theirPhone]}})
+    .then(({data: {findUsers}}) => expect(findUsers.items).toEqual([them]))
+
+  // find us and them by phone and email, make sure they don't duplicate
+  await ourClient
+    .query({query: queries.findUsers, variables: {emails: [ourEmail, theirEmail], phoneNumbers: [theirPhone]}})
+    .then(({data: {findUsers}}) => expect(findUsers.items.sort(cmp)).toEqual([us, them].sort(cmp)))
```
mfogel, line 106: I think these changes are unintentional? I think they might be a product of git merging/branching etc?

integration-testing/tests/find-users.test.js
```diff
@@ -1,10 +1,11 @@
+const moment = require('moment')
 const cognito = require('../utils/cognito')
 const misc = require('../utils/misc')
 const {queries, mutations} = require('../schema')
 const uuidv4 = require('uuid/v4')
 
 const loginCache = new cognito.AppSyncLoginCache()
-jest.retryTimes(1)
+jest.retryTimes(2)
```
mfogel, line 8: I think this change is unintentional also?

integration-testing/tests/find-users.test.js
```diff
@@ -197,3 +186,46 @@ test('Find Users sends cards to the users that were found', async () => {
...
+    expect(before <= self.lastFoundUsers).toBe(true)
+    expect(after >= self.lastFoundUsers).toBe(true)
+  })
+
+  // check another user can't see lastFoundUsers
+  await theirClient.query({query: queries.user, variables: {userId: ourUserId}}).then(({data: {user}}) => {
+    expect(user.userId).toBe(ourUserId)
+    expect(user.lastFoundUsers).toBeNull()
+  })
+})
```
mfogel, line 231: A few thoughts on this test case:

- I would not bother to to initiate all those users so that `findUsers()` returns something. It's ok if `findUsers()` returns an empty list... doesn't matter, `User.lastFoundUsers` should be updated all the same. That will speed up the test and focus it just on just the new field `User.lastFoundUsers`
- At the end of the test, I would run `findUsers()` again to make sure that `User.lastFoundUsers` updates correctly

real-main/app/models/user/model.py
```diff
@@ -470,3 +470,8 @@ def set_dating_status(self, status=None):
 
         self.item = self.dynamo.set_user_dating_status(self.id, status)
         return self
+
+    def update_last_found_time(self, now=None):
```
mfogel, line 474:
```suggestion
    def update_last_found_users_at(self, now=None):
```

real-main/app/models/user/dynamo/base.py
```diff
@@ -149,6 +149,16 @@ def set_user_age(self, user_id, age):
...
         return self.client.update_item(query_kwargs)
 
+    def set_user_last_found_time(self, user_id, now=None):
+        now = now or pendulum.now('utc')
+        query_kwargs = {
+            'Key': self.pk(user_id),
+        }
+        query_kwargs['UpdateExpression'] = 'SET lastFoundUsers = :ps'
+        query_kwargs['ExpressionAttributeValues'] = {':ps': now.to_iso8601_string()}
+
```
mfogel, line 159:
```suggestion
        query_kwargs = {
            'Key': self.pk(user_id),
            'UpdateExpression': 'SET lastFoundUsers = :ps',
            'ExpressionAttributeValues': {':ps': now.to_iso8601_string()},
        }
```

real-main/app/models/user/dynamo/base.py
```diff
@@ -149,6 +149,16 @@ def set_user_age(self, user_id, age):
             query_kwargs['ExpressionAttributeValues'] = {':a': int(age)}
         return self.client.update_item(query_kwargs)
 
+    def set_user_last_found_time(self, user_id, now=None):
```
mfogel, line 152:
```suggestion
    def set_user_last_found_users_at(self, user_id, now=None):
```

mfogel (changes requested):
One final request, this one is my fault.

Can you please rename the graphql field and dynamo attribute from `lastFoundUsers` to `lastFoundUsersAt` ? Not sure why I didn't realize that this should have an `At` suffix earlier... at least that's a pattern the rest of the api & database follows so far, so it would be good to maintain that pattern here. Thanks.

integration-testing/tests/find-users.test.js
```diff
@@ -197,3 +198,50 @@ test('Find Users sends cards to the users that were found', async () => {
...
+  // Check initialize of lastFoundUsersAt
+  await ourClient.query({query: queries.self}).then(({data: {self}}) => {
+    expect(self.lastFoundUsersAt).toBeNull()
+  })
+
+  // Run the findUsers Query
+  let before = moment().toISOString()
+  await ourClient
+    .query({query: queries.findUsers, variables: {emails: [ourEmail]}})
+    .then(({data: {findUsers}}) => expect(findUsers.items).toEqual([us]))
```
mfogel, line 223:
```suggestion
    .then(({data: {findUsers}}) => expect(findUsers.items.map((i) => i.userId)).toEqual([ourUserId]))
```

integration-testing/tests/find-users.test.js
```diff
@@ -197,3 +198,50 @@ test('Find Users sends cards to the users that were found', async () => {
...
+  await theirClient.query({query: queries.user, variables: {userId: ourUserId}}).then(({data: {user}}) => {
+    expect(user.userId).toBe(ourUserId)
+    expect(user.lastFoundUsersAt).toBeNull()
+  })
+
+  // Call findUsers again and check the lastFoundUsersAt is updated correctly
+  await ourClient
+    .query({query: queries.findUsers, variables: {emails: [ourEmail]}})
+    .then(({data: {findUsers}}) => expect(findUsers.items).toEqual([us]))
+
```
mfogel, line 243:
```suggestion
    .then(({data: {findUsers}}) => expect(findUsers.items.map((i) => i.userId)).toEqual([ourUserId]))
  await misc.sleep(2000)
```

real-main/app/models/user/dynamo/base.py
```diff
@@ -149,6 +149,16 @@ def set_user_age(self, user_id, age):
...
         return self.client.update_item(query_kwargs)
 
+    def set_user_last_found_users_at(self, user_id, now=None):
+        now = now or pendulum.now('utc')
+        query_kwargs = {
+            'Key': self.pk(user_id),
+            'UpdateExpression': 'SET lastFoundUsersAt = :ps',
+            'ExpressionAttributeValues': {':ps': now.to_iso8601_string()},
+        }
+
```
mfogel, line 159:
```suggestion

```

real-main/schema.graphql
```diff
@@ -495,6 +495,9 @@ type User {
   #   - a user that has blocked caller or that caller has blocked
   #   - a user with whom caller has not exchanged direct messages
   directChat: Chat
+
+  # last timestamp which call findUsers
```
mfogel, line 499:
```suggestion
  # when user last called findUsers()
```

real-main/app/handlers/appsync/handlers.py
```diff
@@ -1422,4 +1422,5 @@ def find_users(caller_user, arguments, **kwargs):
         raise ClientException('Cannot submit more than 100 combined emails and phoneNumbers')
 
     user_ids = user_manager.find_users(caller_user, emails=emails, phones=phones)
+    caller_user.update_last_found_users_at(pendulum.now('utc'))
```
mfogel, line 1425: In my opinion, either this

```suggestion
    caller_user.update_last_found_users_at(now=pendulum.now('utc'))
```

or this

```suggestion
    caller_user.update_last_found_users_at()
```

would be better. The reason is that the parameter accepted by `update_last_found_users()` is a keyword argument.

Although python allows you to provide values for keyword arguments using their position, I think using that is generally bad... python 3 added a way to explicitly forbid this but it makes all your function definitions three characters longer, so I didn't use it in this project.

To me, this it's an error in python's design that you can pass values  to keyword arguments using their position.

The style guide here: https://python-docs.readthedocs.io/en/latest/writing/style.html talks a bid about this if you want to read more on it.

real-main/app/models/user/model.py
```diff
@@ -470,3 +470,8 @@ def set_dating_status(self, status=None):
 
         self.item = self.dynamo.set_user_dating_status(self.id, status)
         return self
+
+    def update_last_found_users_at(self, now=None):
+        now = now or pendulum.now('utc')
+        self.dynamo.set_user_last_found_users_at(self.id, now)
```
mfogel, line 476:
```suggestion
        self.dynamo.set_user_last_found_users_at(self.id, now=now)
```

For the same reason as the previous comment.

mfogel (changes requested):
Thanks for the changes. This new round of suggestions is more minor stuff but things I think are worth addressing before merging in.

Thank you!

integration-testing/tests/find-users.test.js
mfogel: This suggestion, along with the one above it, allow the `us` variable to be eliminated, as well as `ourUsername`.

PR #71, opened 2020-09-30
Bug:Steal user email/phoneNumber exception

zkhin:
Fixes: https://trello.com/c/nKpMIivz/24-user-attempting-to-steal-an-email-address-from-another-user-results-in-errors-and-inconsistent-backend-state


real-main/app/models/user/model.py
```diff
@@ -350,6 +351,12 @@ def start_change_contact_attribute(self, attribute_name, attribute_value):
             names['cognito']: attribute_value,
             f'custom:unverified_{names["short"]}': attribute_value,
         }
+
+        # verify that new attribtue value is not used by other
+        attribute_dynamo = UserContactAttributeDynamo(self.dynamo, f'user{names["dynamo"].title()}')
```
mfogel: Nice job figuring this out. There's some improvements we can do.

Rather than initiating  a new copy of the `UserContactAttributeDynamo` library each time this method is called, on each `User` model instance, it would be better if we re-use the one that the `UserManager` already has. All the dynamo libraries (in fact all the python code except the models themselves - in `model.py`) are designed to be state-less, hence they should be shared like this as much as possible. Here's what I would do.

- add two new kwargs to `User.__init__()`: `email_dynamo` and `phone_number_dynamo`. Save those as `self.email_dynamo` and `self.phone_number_dynamo`
- make changes to `UserManager.init_user()` so that `UserManager.email_dyanmo` and `UserManager.phone_number_dynamo` are both based in to `User.__init__()`
- use `getattr()` here to select the right library: either `self.email_dynamo` or `self.phone_number_dynamo`

integration-testing/tests/users/steal-user-email-phone.test.js
```diff
@@ -0,0 +1,56 @@
...
+const cognito = require('../../utils/cognito')
+
+const loginCache = new cognito.AppSyncLoginCache()
+jest.retryTimes(1)
+
+beforeAll(async () => {
+  const ourPhone = '+15105551000'
+  const theirPhone = '+15105551111'
+  loginCache.addCleanLogin(await cognito.getAppSyncLogin(ourPhone))
+  loginCache.addCleanLogin(await cognito.getAppSyncLogin(theirPhone))
```
mfogel, line 12: Two thoughts here:

- `+15105551000` is already in use by another integration test. We have to use unique phone numbers in each test file, so that they can be run in parallel. So I would change this to some other valid phone number that is not yet in use by our integration tests.
- I wouldn't bother to declare `ourPhone` and `theirPhone` here... I generally stay away from declaring variables just as documentation unless it's something seriously complicated. So to me, doing `loginCache.addCleanLogin(await cognito.getAppSyncLogin('+1...'))` would be better.

integration-testing/tests/users/steal-user-email-phone.test.js
```diff
@@ -0,0 +1,56 @@
...
+})
+beforeEach(async () => await loginCache.clean())
+afterAll(async () => await loginCache.reset())
+afterEach(async () => {})
```
mfogel, line 16:
```suggestion
```

No need for the empty `afterEach()`, can be removed.

integration-testing/tests/users/steal-user-email-phone.test.js
```diff
@@ -0,0 +1,56 @@
...
+const startChangeUserPhoneNumber = gql`
+  mutation StartChangeUserPhoneNumber($phoneNumber: AWSPhone!) {
+    startChangeUserPhoneNumber(phoneNumber: $phoneNumber) {
+      userId
+      username
+      email
+      phoneNumber
+    }
+  }
+`
```
mfogel, line 38: I would put these in `schema/mutations.js`

integration-testing/tests/users/steal-user-email-phone.test.js
```diff
@@ -0,0 +1,56 @@
...
+
+test('Enable, disable dating as a BASIC user, privacy', async () => {
+  const {email: ourEmail, client: ourClient} = await loginCache.getCleanLogin()
+  const {client: theirClient} = await loginCache.getCleanLogin()
+  let ourPhoneNumber
+
+  await ourClient.query({query: queries.self}).then(({data: {self: user}}) => {
+    ourPhoneNumber = user.phoneNumber
+  })
```
mfogel, line 47:
```suggestion

  const ourPhoneNumber = await ourClient.query({query: queries.self}).then(({data: {self: user}}) => user.phoneNumber)
```

mfogel (changes requested):
Nice job figuring this out in an email/phone agnostic way. And also thanks for adding an integration test!

Let me know if you thoughts/questions on these comments, cheers.

integration-testing/tests/users/steal-user-email-phone.test.js
zkhin: I used this code instead of above but I am getting the `null` value for the phoneNumber
```
loginCache.addCleanLogin(await cognito.getAppSyncLogin('+15105551000'))
loginCache.addCleanLogin(await cognito.getAppSyncLogin('+15105551111'))
```
But if I use old code, it works fine... weird...

mfogel: Yeah, what's probably going on is there is already a user in your dynamo & cognito with that phone number. The users in the integration test suite are linked to the line number of the file that creates them. So since the line numbers change with this change, then the test is unable to create another user with that same phone number (since one already exists).

We avoid this with emails by generating valid emails on-the-fly. Can't do that with phone numbers... at least I don't know how. It's not as simple as picking a random number.

What you can do to fix your test env:

- run the script `integration-testing/clean/clear-dynamodb-table`
- see if that fixes it
- if not, run that script again and then also run `integration-testing/clean/clear-user-pool` (takes a few minutes to run)

zkhin: Stucked while running `cleaer-user-pool` script
image

real-main/app/models/user/model.py
```diff
@@ -350,6 +356,11 @@ def start_change_contact_attribute(self, attribute_name, attribute_value):
             names['cognito']: attribute_value,
             f'custom:unverified_{names["short"]}': attribute_value,
         }
+
+        # verify that new attribtue value is not used by other
+        if self.email_dynamo.get(attribute_value) or self.phone_number_dynamo.get(attribute_value):
```
mfogel, line 361: This works but as I imagine you know, it will query email addresses for a phone number, which is a waste of a query. Here's one way to avoid that useless query.

First, change the definition of `CONTACT_ATTRIBUTE_NAMES` like so:

```py
CONTACT_ATTRIBUTE_NAMES = {
    'email': {
        'short': 'email',
        'cognito': 'email',
        'dynamo_attr': 'email',
        'dynamo_client': 'email_dynamo',
    },
    'phone': {
        'short': 'phone',
        'cognito': 'phone_number',
        'dynamo_attr': 'phoneNumber',
        'dynamo_client': 'phone_number_dynamo',
    },
}
```

then this part here becomes

```suggestion
        contact_attr_dyamo = getattr(self, names[attribute_name])
        if contact_attr_dynamo.get(attribute_value):
```

Also, existing code of `names['dynamo']` will need to be changed to `names['dynamo_attr']`

PR #72, opened 2020-10-01
Change findUsers to findContacts

zkhin:
fixes: https://trello.com/c/yAmly0Er/25-change-queryfindusers-into-queryfindcontacts


real-main/app/models/user/manager.py
```diff
@@ -504,27 +504,48 @@ def on_user_delete_delete_cognito(self, user_id, old_item):
...
+                phone_contacts[phone] = contact['contactId']
+
+        user_ids_mapped_emails = self.email_dynamo.batch_get_user_ids_attr_mapped(email_contacts.keys())
+        user_ids_mapped_phones = self.phone_number_dynamo.batch_get_user_ids_attr_mapped(phone_contacts.keys())
+        user_ids_mapped_attrs = {**user_ids_mapped_emails, **user_ids_mapped_phones}
+        contact_ids_mapped_attrs = {**email_contacts, **phone_contacts}
+
+        user_ids_mapped_contact_ids = {}
+        for attr in user_ids_mapped_attrs.keys():
+            user_id = user_ids_mapped_attrs[attr]
```
mfogel, line 530:
```suggestion
        for attr, user_id in user_ids_mapped_attrs.items():
```

real-main/app/models/user/manager.py
```diff
@@ -504,27 +504,48 @@ def on_user_delete_delete_cognito(self, user_id, old_item):
...
+        for contact in contacts:
+            for email in contact.get('emails', []):
+                email_contacts[email] = contact['contactId']
+            for phone in contact.get('phones', []):
+                phone_contacts[phone] = contact['contactId']
+
+        user_ids_mapped_emails = self.email_dynamo.batch_get_user_ids_attr_mapped(email_contacts.keys())
+        user_ids_mapped_phones = self.phone_number_dynamo.batch_get_user_ids_attr_mapped(phone_contacts.keys())
+        user_ids_mapped_attrs = {**user_ids_mapped_emails, **user_ids_mapped_phones}
+        contact_ids_mapped_attrs = {**email_contacts, **phone_contacts}
```
mfogel, line 526: Naming suggestion, take it or leave it as you like. This is personal preference, but I would find this more clear with variable names like

```
email_to_user_id = ...
phone_to_user_id = ...
conatct_attr_to_user_id = ...
```
etc

real-main/app/models/user/manager.py
```diff
@@ -504,27 +504,48 @@ def on_user_delete_delete_cognito(self, user_id, old_item):
...
+            contact_id = contact_ids_mapped_attrs[attr]
+            user_ids_mapped_contact_ids[contact_id] = user_id
+
+        # sort dict by key(contact_id)
+        user_ids_mapped_contact_ids = {
+            k: user_ids_mapped_contact_ids[k] for k in sorted(user_ids_mapped_contact_ids)
+        }
+
+        for contact_id in user_ids_mapped_contact_ids.keys():
+            user_id = user_ids_mapped_contact_ids[contact_id]
```
mfogel, line 540:
```suggestion
        for contact_id, user_id in user_ids_mapped_contact_ids.items():
```

real-main/app/models/user/manager.py
```diff
@@ -504,27 +504,48 @@ def on_user_delete_delete_cognito(self, user_id, old_item):
...
-        phones = phones or []
 
-        user_ids_from_emails = self.email_dynamo.batch_get_user_ids(emails)
-        user_ids_from_phones = self.phone_number_dynamo.batch_get_user_ids(phones)
-        user_ids = list(set(user_ids_from_emails + user_ids_from_phones))  # uniquify
+        found_contacts = []
+        email_contacts, phone_contacts = {}, {}
+        for contact in contacts:
+            for email in contact.get('emails', []):
+                email_contacts[email] = contact['contactId']
```
mfogel, line 519:
```suggestion
            email_contacts.update({email: contact['contactId'] for email in contact.get('emails, [])})
```

Same thing can be done for phone, of course.

mfogel (changes requested):
Some minor performance improvements, python niceties

real-main/app/models/user/manager.py
```diff
@@ -504,27 +504,48 @@ def on_user_delete_delete_cognito(self, user_id, old_item):
...
+        user_ids_mapped_contact_ids = {}
+        for attr in user_ids_mapped_attrs.keys():
+            user_id = user_ids_mapped_attrs[attr]
+            contact_id = contact_ids_mapped_attrs[attr]
+            user_ids_mapped_contact_ids[contact_id] = user_id
+
+        # sort dict by key(contact_id)
+        user_ids_mapped_contact_ids = {
+            k: user_ids_mapped_contact_ids[k] for k in sorted(user_ids_mapped_contact_ids)
+        }
```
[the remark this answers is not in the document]
zkhin: Yes, I sorted to figure out the unit test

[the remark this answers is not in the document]
zkhin: Can you suggest a better idea to figure out the spec without sorting in the backend?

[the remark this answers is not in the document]
mfogel: I would actually consider changing the response format of `UserManager.find_contacts()` to just return a python dictionary of contact_id => user_id

In the handler in `app/handers/appsync/handlers.py` I would transform the response from `UserManager.find_contacts()` like so:

```py
contact_id_to_user_id = user_manager.find_contacts(...)
return [{'contactId': contact_id, 'userId': user_id} for contact_id, user_id in contact_id_to_user_id.items()]
```

Then I would set up a new pipeline graphql resolver for `FoundContact.user`. This basically just needs to do the same thing as the existing pipeline resolver for `Query.user` it just has to pull the user id from `ctx.source.userId`, rather than `ctx.args.userId`

The reason for this is that I see another place in this implementation that is currently doing `O(n)` queries which I had missed before. I'll point that out.

As far as I remember, none of our graphql resolver do `O(n)` queries yet. In general, good API design shouldn't require ever doing `O(n)` queries in any resolver, IMHO

real-main/app/models/user/manager.py
```diff
@@ -504,27 +504,42 @@ def on_user_delete_delete_cognito(self, user_id, old_item):
...
+        contact_id_to_user_id = {}
+        for attr, user_id in contact_attr_to_user_id.items():
+            contact_id = contact_attr_to_contact_id[attr]
+            contact_id_to_user_id[contact_id] = user_id
+
+        # sort dict by key(contact_id)
+        contact_id_to_user_id = {k: contact_id_to_user_id[k] for k in sorted(contact_id_to_user_id)}
+
+        for contact_id, user_id in contact_id_to_user_id.items():
+            found_contacts.append({'contactId': contact_id, 'user': self.get_user(user_id)})
```
mfogel, line 535: `O(n)` queries here!

If you apply the changes I mentioned in https://github.com/real-social-media/backend/pull/72#discussion_r498913555 I think this will disappear.

real-main/mapping-templates/FoundContact.user/after.response.vtl
```diff
@@ -0,0 +1,22 @@
+#if ($ctx.error)
+  $util.error($ctx.error.message, $ctx.error.type)
+#end
+
+#if ($ctx.prev.result.isEmpty())
+  #return
```
mfogel, line 6: This `#return` right here is a problem, because in the GQL schema specifies this as non-null: `User!`. I should have foreseen this.

I can see two ways to fix this. One would be to move the pipeline resolver from `FoundContact.user` to `Query.foundContacts`. Another would be to change `FoundContact.user` so it can be null. The second one is slightly less performant but it should be easier to implement so let's go with it.

I'm going to suggest a series of changes to this PR to get there, coming soon.

real-main/schema.graphql
```diff
@@ -929,3 +924,14 @@ type CognitoTokens {
...
+
+input ContactInput {
+  contactId: ID!
+  emails: [AWSEmail!]
+  phones: [AWSPhone!]
+}
+
+type FoundContact {
+  contactId: ID!
+  user: User!
```
mfogel, line 936:
```suggestion
  user: User
```

Previously, `Query.findContacts()` only returned `FoundContact` for users that were found. The changes I suggest now will change this so that a `FoundContact` is always returned, however, `FoundContact.user` will be `null` if no user was found.

real-main/mapping-templates/FoundContact.user/after.response.vtl
```diff
@@ -0,0 +1,22 @@
...
+#if ($user.blockerStatus == 'BLOCKING')
+  ## we are blocked! we see only an absolute minimum profile
+  #set ($user = {
+    'userId': $user.userId,
+    'username': $user.username,
+    'blockedStatus': $user.blockedStatus,
+    'blockerStatus': $user.blockerStatus,
+    'followedStatus': $user.followedStatus,
+    'followerStatus': $user.followerStatus
+  })
```
mfogel, line 19: Let's just hide users that have blocked the caller completely here. Thus it will appear to the caller (the user that's been blocked) that the other user doesn't exist.

```suggestion
  #return
```

real-main/mapping-templates/FoundContact.user/after.response.vtl
zkhin: You mean this line should be changed to `user: User`, right?
https://github.com/real-social-media/backend/blob/29dca43c09dda50d41cd88d63ba7cccae036a366/real-main/schema.graphql#L936

real-main/app/handlers/appsync/handlers.py
```diff
@@ -1408,19 +1408,15 @@ def lambda_server_error(caller_user_id, arguments, context=None, **kwargs):
...
-        raise ClientException('Cannot submit more than 100 combined emails and phoneNumbers')
+    if len(contacts) > 100:
+        raise ClientException('Cannot submit more than 100 contact inputs')
 
-    user_ids = user_manager.find_users(caller_user, emails=emails, phones=phones)
-    caller_user.update_last_found_users_at(now=pendulum.now('utc'))
-    return {'items': user_ids}
+    contact_id_to_user_id = user_manager.find_contacts(caller_user, contacts=contacts)
+    caller_user.update_last_found_contacts_at(now=pendulum.now('utc'))
+    return [{'contactId': contact_id, 'userId': user_id} for contact_id, user_id in contact_id_to_user_id.items()]
```
mfogel, line 1422:
```suggestion
    contact_ids = [contact['contactId'] for contact in contacts]
    return [{'contactId': contact_id, 'userId': contact_id_to_user_id.get(contact_id)} for contact_id in contact_ids]
```

real-main/mapping-templates/FoundContact.user/before.request.vtl
```diff
@@ -0,0 +1 @@
+$util.toJson([$ctx.source.userId])
```
mfogel, line 1:
```suggestion
#if ($util.isNull($ctx.source.userId))
  #return
#end
$util.toJson([$ctx.source.userId])
```

mfogel (changes requested):
Ok, so the main change here is that the response of `Query.findContacts()` should contain an entry for each `contactId` the client sends.

If the backend found the user, then `FoundContact.user` contains that user.
If the backend did not find the user (or the user was found, but is blocking the caller), then `FoundContact.user` is null.

The integration tests will need to be updated to match this change. (I didn't spec out exactly were they have to change, I haven't read them through closely yet).

Let me know if you have questions, thank you

mfogel (commented):
I added a few comments on the integration tests primarily.

I would add an integration test. One that check to make sure that users that have blocked the query are not returned by `Query.findContacts`. Here's one way to do that:
- two users, us and them
- verify we can find them using their email
- they block us
- verify we can no longer find them using their email

I haven't reviewed the python unit tests yet... I will do so tomorrow.

real-main/app/handlers/appsync/handlers.py
```diff
@@ -1416,19 +1416,18 @@ def lambda_server_error(caller_user_id, arguments, context=None, **kwargs):
...
-    caller_user.update_last_found_users_at(now=pendulum.now('utc'))
-    return {'items': user_ids}
+def find_contacts(caller_user, arguments, **kwargs):
+    contacts = arguments['contacts']
+
+    if len(contacts) > 100:
+        raise ClientException('Cannot submit more than 100 contact inputs')
+
+    contact_id_to_user_id = user_manager.find_contacts(caller_user, contacts=contacts)
+    caller_user.update_last_found_contacts_at(now=pendulum.now('utc'))
```
mfogel, line 1429: Even though I don't know if these actually do throw a `UserException`, I think it's still better to wrap them in this try-catch block because they may be changed in the future to do so.

Also, `contacts` is no longer a kwarg (it's just a normal positional argument) so I wouldn't use the `contacts=contacts` syntax on it, I would just pass in `contacts`.

```suggestion
    try:
        contact_id_to_user_id = user_manager.find_contacts(caller_user, contacts)
        caller_user.update_last_found_contacts_at(now=pendulum.now('utc'))
    except UserException as err:
        raise ClientException(str(err)) from err
        
```

mfogel (changes requested):
This is very close - just the one suggestion here and then also, because in the mapping template blocked users are handled differently there should be a test exercising that behavior.

I would add an integration test that goes like this:

- two users, us and them
- verify we can find them using their email
- they block us
- verify we can no longer find them using their email

integration-testing/tests/find-contacts.test.js
```diff
@@ -0,0 +1,314 @@
...
+test('Find contacts can handle duplicate emails', async () => {
+  const {client, userId, email, username} = await loginCache.getCleanLogin()
+  await misc.sleep(2000)
+  const contacts = [{contactId: 'contactId', emails: [email, email]}]
+  await client.query({query: queries.findContacts, variables: {contacts}}).then(({data: {findContacts}}) => {
+    expect(findContacts).toHaveLength(1)
+    expect(findContacts[0].contactId).toBe('contactId')
+    expect(findContacts[0].user.userId).toBe(userId)
+    expect(findContacts[0].user.username).toBe(username)
+    expect(findContacts[0].user.photo).toBeTruthy()
```
mfogel, line 37: This line is failing for me when I run these tests. `user.photo` is `null`. I'm not sure why. Does it fail for you? All the other integration tests are passing.

Side note: This assertion on `user.photo` is useful - as demonstrated by the fact that it's failing. I'm not sure about all the assertions on `user.username`, in every test. I might do assertions on `user.username` in _one_ test (perhaps something like 'test extra user fields'). I might also add assertions on a handful of other fields in that test. But I wouldn't bother to do those extra assertions in every test.

zkhin: Hmm... it's passing all the tests on my side...strange.
image

mfogel: crazy. thank you - I will investigate on my side to figure out why my dev env is failing.

mfogel: Ok, I figured out what's going on. I think this test will fail for you in your dev env: https://github.com/real-social-media/backend/blob/master/integration-testing/tests/users/create-user/placeholder-photo.test.js#L21. Can you verify that it fails for you?

There is an unfortunate coupling between the backend project and the `themes` project (which is used to hold some images & files used by the frontend). We have an unused feature where when a new user is created, the backend scans a directory in the `themes` project that is full of default profile pictures, and selects a random one, and sets that as the new user's profile picture. However, we only ever added one default profile picture to that directory... so there are much simpler implementations possible to accomplish that than what now exists. And that would avoid this awkward coupling.

The result is that the integration tests are currently set up to expect that directory full of default profile pictures to be empty. I think this test, on this line, assumes that directory is _not_ empty. And I suspect that directory in your dev environment is not empty.

As such - could we just change this from `user.photo` to something else? Like maybe `user.subscriptionLevel` or `user.signedUpAt` ? Those attributes don't run into this gotcha.

PR #73, opened 2020-10-05
Strip timezone information from dateOfBirth

zkhin:
Fixes: https://trello.com/c/bWUPUymr/28-dynamo-field-userdateofbirth-should-not-have-timezone-information

### DIFF
--- a/integration-testing/tests/users/properties.test.js
+++ b/integration-testing/tests/users/properties.test.js
@@ -79,6 +79,25 @@ describe('Read and write properties our our own profile', () => {
     expect(resp.data.user.fullName).toBeNull()
   })
 
+  test('Invalid dateOfBirth contains timezone information', async () => {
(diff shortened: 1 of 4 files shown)


mfogel (commented):
This looks ok so far, but here are some other date formats this needs to handle: (from https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html)

- `1970-01-01-07:00`
- `1970-01-01+05:30`

Another way to implement this occurred to me just now:

- add another validator to `app/handlers/appsync/validation.py`: `def validate_date_of_birth(date_of_birth)`
- have that raise a `ClientException` if there is timezone information present

that would probably be a slightly 'better' implementation than the current one of just ignoring timezone information... because it's more explicit to the client about exactly what's going on. But if you would rather continue with the one you've already done (just make it work for those additional date formats above) that is ok too, I will merge either implementation.

PR #76, opened 2020-10-15
Block chat from rejected approved matches, unless confirmed

zkhin:
Fixes: https://trello.com/c/LvZScK1X/21-block-chats-from-rejected-approved-matches-unless-confirmed

### DIFF
--- a/real-main/app/models/chat/manager.py
+++ b/real-main/app/models/chat/manager.py
@@ -89,7 +89,7 @@ class ChatManager(FlagManagerMixin, ViewManagerMixin, ManagerBase):
 
         # can't add a chat if the match status is rejected/approved
         if not self.validate_dating_match_chat(created_by_user_id, with_user_id):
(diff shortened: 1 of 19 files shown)


real-main/app/models/chat/manager.py
```diff
@@ -83,6 +87,10 @@ def add_direct_chat(self, chat_id, created_by_user_id, with_user_id, now=None):
         if with_user.status != UserStatus.ACTIVE:
             raise ChatException(f'Cannot open direct chat with user with status `{with_user.status}`')
 
+        # can't add a chat if the match status is rejected/approved
+        if not self.validate_dating_match_chat(created_by_user_id, with_user_id):
+            raise ChatException('Cannot open direct chat with user without confirmed match')
```
real-ian, line 92: let's make this "Can't chat user viewed on dating unless it's a match"

PR #89, opened 2020-11-09
Support png format

zkhin:
Fixes: https://trello.com/c/4wRXsXq8/41-investigate-why-png-uploads-are-failing-processing-they-should-be-unverified-but-not-fail-to-upload-entirely


real-ian (commented):
please delete the commented test. Then we can merge :)

PR #91, opened 2020-11-10
Update lastDeflatedAt even if the score is zero

zkhin:
Fixes: https://trello.com/c/6O1o9xcJ/52-investigate-cloudwatch-error-addscore-method-in-trendingdynamo


real-ian (commented):
@zkhin are you sure this won't lead to posts getting deflated to have a negative score? 

Also let's confirm that this won't lead to forever trending posts (make sure posts which have a score of 0 are still being removed from the trending index). 

I think this is the correct solution, I just want to tripple check! Thank you!

PR #107, opened 2020-12-11
Detect bad words in downcase

zkhin:
Fixes: https://trello.com/c/Z2bCda4h/87-check-to-ensure-new-comments-chat-messages-are-checked-for-bad-words-in-downcase-format-this-way-we-dont-need-to-add-bad-words-i


real-ian:
elegant, perfect :) thanks dude

PR #112, opened 2020-12-14
Users who are banned should not be able to sign up again

zkhin:
Fixes: https://trello.com/c/GSOXY60r/72-users-who-are-banned-should-not-be-able-to-sign-up-again-that-deviceid-email-and-phone-number-should-be-blacklisted-from-signup


real-ian:
Hey @zkhin the check if a user is already banned exists in the user model, but the user would still be created in the identity pool if they sign up with a 3rd party (e.g google) right? Just thinking about how we would unban someone if they were falsely banned. From this I think we would need to:

1) delete any record of their user in the user pool & identity pool
2) delete their device_uuid from the banned devices in dynamo ?

real-ian:
Let's test this all in staging before any merge to prod

PR #121, opened 2020-12-16
Add s3 event to trigger detecting bad words updated

zkhin:
Fixes: https://trello.com/c/DtUS7e4W/95-run-the-detecting-bad-words-script-when-the-badword-list-is-updated-using-s3-event

### DIFF
--- a/real-main/serverless.yml
+++ b/real-main/serverless.yml
@@ -656,7 +656,12 @@ functions:
     layers:
       - ${cf:real-${self:provider.stage}-lambda-layers.PythonRequirementsLambdaLayer}
     events:
-      - schedule: 'rate(1 day)'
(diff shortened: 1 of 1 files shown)


real-ian:
Hey @zkhin I wonder if the "s3 object created" event triggers when we edit the file live, or do a diff like S3 file sync. 

A quick google makes it appear that it will work, but wondering if you have seen anything: https://stackoverflow.com/questions/39500813/does-aws-s3-have-a-concept-of-files-being-updated/39501249#:~:text=No%2C%20there%20is%20no%20concept,through%20a%20PUT%20object%20request.&text=An%20S3%20bucket%20notification%20configured,can%20execute%20a%20Lambda%20function.

PR #155, opened 2021-02-24
New redeem promotion mutation

zkhin:
https://trello.com/c/pO9uDFqT/133-promotions


real-ian:
Hey @zkhin  one thing I don't see in the test case, what happens when you try to redeem twice the same code? It should return an error.

real-ian:
Hey @zkhin  one thing I don't see in the test case, what happens when you try to redeem twice the same code? It should return an error.

PR #193, opened 2021-05-07
Performnace: Add a bunch of appsync caching

mfogel:
This PR sets up short caching (1 second) on nearly all query resolvers.

The idea is that the frontend ends up requesting a lot of duplicate data within the same query. For example, a query to `Query.trendingPosts` might be for 54 posts but those 54 posts all belong to say 10 total users. So resolvers that are executed for the fields on `Query.trendingPosts.items.X.ownedBy` will hopefully hit the cache part of the time, even though the TTL is only 1 second.

We could raise TTL above 1 second, but it may start to make Mutations complicated for the frontend since changes to state may not appear until the TTL expires.


mfogel:
I ran the integration tests against this PR, they pass. However, that doesn't guarantee there won't be issues in the app once this is merged.

PR #208, opened 2021-05-12
Add dating matched card

zkhin:
https://real-social.atlassian.net/browse/BE-151

### DIFF
--- a/real-main/app/models/card/templates.py
+++ b/real-main/app/models/card/templates.py
@@ -19,7 +19,7 @@ class CardTemplate:
 class ChatCardTemplate(CardTemplate):
 
     action = 'https://real.app/chat/'
-    notify_user_after = pendulum.duration(seconds=5)
+    notify_user_after = pendulum.duration(minutes=5)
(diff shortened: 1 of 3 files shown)


real-main/app/models/card/templates.py
```diff
@@ -19,7 +19,7 @@ def __init__(self, user_id):
 class ChatCardTemplate(CardTemplate):
 
     action = 'https://real.app/chat/'
-    notify_user_after = pendulum.duration(seconds=5)
+    notify_user_after = pendulum.duration(minutes=5)
```
mfogel, line 22: Was this change intentional? I think this will affect how long before someone gets notified they have an unread message in chat

zkhin: Yeah, this was changed to `5 seconds` (origin was 5 minutes) for testing the APNS issue and forgot to roll back the duration.

mfogel: ok, cool

PR #209, opened 2021-05-14
Add new follower card

zkhin:
https://real-social.atlassian.net/browse/BE-152

### DIFF
--- a/real-main/app/handlers/dynamo/handlers.py
+++ b/real-main/app/handlers/dynamo/handlers.py
@@ -240,7 +240,7 @@ register(
 register(
     'user',
     'follower',
-    ['INSERT'],
+    ['INSERT', 'MODIFY'],
     user_manager.on_user_new_followers_sync_card,
     {'followStatus': FollowStatus.NOT_FOLLOWING},
 )
--- a/real-main/app/handlers/dynamo/handlers.py
+++ b/real-main/app/handlers/dynamo/handlers.py
@@ -237,6 +237,13 @@ register(
     user_manager.sync_follow_counts_due_to_follow_status,
     {'followStatus': FollowStatus.NOT_FOLLOWING},
 )
+register(
+    'user',
+    'follower',
+    ['INSERT'],
+    user_manager.on_user_new_followers_sync_card,
+    {'followStatus': FollowStatus.NOT_FOLLOWING},
+)
 register('user', 'profile', ['INSERT'], user_manager.on_user_add_delete_user_deleted_subitem)
 register(
     'user',
--- a/real-main/app/models/user/manager.py
+++ b/real-main/app/models/user/manager.py
@@ -13,13 +13,17 @@ from app.clients import SesClient
 from app.mixins.base import ManagerBase
 from app.mixins.trending.manager import TrendingManagerMixin
 from app.models.appstore.enums import AppStoreSubscriptionStatus
-from app.models.card.templates import ContactJoinedCardTemplate, UserNewDatingMatchesTemplate
+from app.models.card.templates import (
+    ContactJoinedCardTemplate,
+    NewFollowersCardTemplate,
+    UserNewDatingMatchesTemplate,
+)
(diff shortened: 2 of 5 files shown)


real-main/app/handlers/dynamo/handlers.py
```diff
@@ -237,6 +237,13 @@
     user_manager.sync_follow_counts_due_to_follow_status,
     {'followStatus': FollowStatus.NOT_FOLLOWING},
 )
+register(
+    'user',
+    'follower',
+    ['INSERT'],
```
mfogel, line 243: If I were coding this up, I would make this trigger on `['INSERT', 'MODIFY']`.

Triggering on just `INSERT` works but because of what I would consider to be an implementation detail: we currently don't store these items in the DB when they're in `NOT_FOLLOWING` status.

I would avoid making this implementation depend on that detail unless there's a good reason to. From what I can see it's pretty easy to avoid depending on it.

real-main/app/models/user/manager.py
```diff
@@ -590,6 +594,16 @@ def on_appstore_transaction_add(self, transaction_id, new_item):
...
             self.dynamo.increment_paid_real_so_far(new_item['userId'], price)
 
+    def on_user_new_followers_sync_card(self, user_id, new_item, old_item=None):
+        followed_user = self.get_user(user_id)
+        followed_user_privacy_status = followed_user.item.get('privacyStatus', UserPrivacyStatus.PUBLIC)
+        follow_status = (new_item or {}).get('followStatus', FollowStatus.NOT_FOLLOWING)
+        card_template = NewFollowersCardTemplate(user_id)
+
+        # followed user's status should be public
+        if follow_status == FollowStatus.FOLLOWING and followed_user_privacy_status == UserPrivacyStatus.PUBLIC:
```
mfogel, line 604: If I were writing this up, I would also explicitly check the previous follow status.

```python
old_follow_status = (old_item or {}).get('followStatus', FollowStatus.NOT_FOLLOWING)
# make sure old_follow_status is NOT_FOLLOWING before issuing card
```

real-main/app/models/user/manager.py
```diff
@@ -590,6 +594,16 @@ def on_appstore_transaction_add(self, transaction_id, new_item):
             price = new_item.get('price', Decimal('0'))
             self.dynamo.increment_paid_real_so_far(new_item['userId'], price)
 
+    def on_user_new_followers_sync_card(self, user_id, new_item, old_item=None):
+        followed_user = self.get_user(user_id)
+        followed_user_privacy_status = followed_user.item.get('privacyStatus', UserPrivacyStatus.PUBLIC)
+        follow_status = (new_item or {}).get('followStatus', FollowStatus.NOT_FOLLOWING)
```
mfogel, line 600: I think new_item is always defined here?

```suggestion
        follow_status = new_item.get('followStatus', FollowStatus.NOT_FOLLOWING)
```

real-main/app/models/user/manager.py
```diff
@@ -590,6 +594,16 @@ def on_appstore_transaction_add(self, transaction_id, new_item):
             price = new_item.get('price', Decimal('0'))
             self.dynamo.increment_paid_real_so_far(new_item['userId'], price)
 
+    def on_user_new_followers_sync_card(self, user_id, new_item, old_item=None):
+        followed_user = self.get_user(user_id)
+        followed_user_privacy_status = followed_user.item.get('privacyStatus', UserPrivacyStatus.PUBLIC)
```
mfogel, line 599: For performance, it would be better to do these operations only after checking the `new_follow_status` and `old_follow_status`, since they trigger a query to dynamo.

So I would do something like:

```python
def on_user_new_followers_sync_card(self, user_id, new_item, old_item=None):
    new_follow_status = ...
    old_follow_status = ...
    if (new_follow_status != FOLLOWING or old_follow_status != NOT_FOLLOWING) return
    user = self.get_user(user_id)
    # verify user is public user before creating card
```

PR #211, opened 2021-05-18
ID analyzer

zkhin:
https://real-social.atlassian.net/browse/BE-138


real-main/app/clients/id_verification.py
```diff
@@ -56,3 +61,20 @@ def verify_id(self, user_id, frontside_image, country, id_type, mime_type):
             raise Exception(
                 f'Unable to parse response from jumio verification service with body: `{resp.text}`'
             ) from err
+
+    def verify_id_with_id_analyzer(self, frontside_image):
+        if not self.id_analyzer_api_key:
+            return
```
mfogel, line 67: I think this check could hide errors if the entries in the SecretsManager are misconfigured? Better to fail early and loudly....

If you're looking for an easy way to disable the service, I would do that either with another property in the secrets manager json or possibly with a special api key of something like `DISABLED`

mfogel (commented):
Nice work. I made a few targeted comments but there is a bigger structural pattern that you may or may not have seen before.

Typically 'clients' are per-service. So in this case there might be a `JumioClient` and an `IdAnalyzerClient`. You can see this in the design of the AWS SDK. You wouldn't typically combine two services into one client.

If I were coding this up, I would probably:
- make two separate client classes, one for each service. They would present very similar public apis to the rest of the code base. As similar as possible.
- I would add unit tests for each of the clients! I would use `requests-mock` to mock the request-response cycle of each service. Here's an example of how to do that: https://github.com/real-social-media/backend/blob/v2021-05-10/real-main/app_tests/clients/test_post_verification.py
- I would add a config variable, probably defined in the `environment` section of  `serverless.yml`, that allows us to choose between
  - enabling jumio: `IDVERIFICATION_SERVICE: jumio`
  - enabling idverify: `IDVERIFICATION_SERVICE: idverify`
  - explicitly disabling both: `IDVERIFICATION_SERVICE: disabled`

real-main/app/clients/id_verification.py
```diff
@@ -56,3 +61,20 @@ def verify_id(self, user_id, frontside_image, country, id_type, mime_type):
...
+        # https://developer.idanalyzer.com/coreapi_reference.html
+        payload = {'apikey': self.id_analyzer_api_key, 'file_base64': frontside_image}
+        r = requests.post(IDANALYZER_API_URL, data=payload)
+        result = r.json()
+
+        if 'error' in result:
```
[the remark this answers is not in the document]
zkhin: Yes, it always returns 200 response with error code.

```
code: 200
{
    "error": {
        "message": "Document not recognized",
        "code": 9
    },
    "executionTime": 1.1604459285736084,
    "responseID": "2db13335fed442fc15aa0c123e69f821"
}
```

real-main/app/handlers/appsync/handlers.py
```diff
@@ -1518,23 +1519,14 @@ def flag_chat_message(caller_user, arguments, **kwargs):
...
-    image_type = arguments['imageType']
-
     try:
-        caller_user.verify_id_document(
-            frontside_image=frontside_image,
-            country=country,
-            id_type=id_type,
-            image_type=image_type,
-        )
+        caller_user.verify_id_document_with_id_analyzer(frontside_image=frontside_image)
```
mfogel, line 1529: So I don't want to add to much to this task, but if I were coding this up, I would try to do the actual verification with the external id verification service asynchronously using dynamo streams. So that would involve writing some changes to dynamo to record the frontend's ID verification request, then actually doing the call to the IDVerification service from the dynamo stream handler.

This is partly for performance, but more so for reliability and error isolation. Like if IdVerify goes down, and we're accessing it synchronously, then all of the sudden we're sending server errors to the frontend, which it should be able to handle but probably isn't a scenario they test everyday...

Feel free to ignore this comment if you'd rather move on to the next task 🙂

mfogel (changes requested):
Nice work @zkhin , let me know if you have questions/thoughts

real-main/app/handlers/appsync/handlers.py
zkhin: That makes sense and it will be better to use a dynamo stream handler.
If you don't mind, I would like to create a new task for this refactoring and will come back again later.

mfogel (approved):

real-main/app/clients/id_verification.py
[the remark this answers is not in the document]
mfogel: 👍

PR #218, opened 2021-05-31
BE-160: Ads

mfogel:
https://real-social.atlassian.net/jira/software/projects/BE/boards/2?selectedIssue=BE-160

Highlights:

- new user setting `User.adsDisabled`, defaults to `False`
- posts get a new enum `Post.adStatus`, defaults to `NOT_AD`
- GQL mutation `addPost()` has new options `isAd`, `adPayment`, and `adPaymentPeriod`
- posts that are ads continue to do everything normal posts do
- users that have `adsDisabled=True` will never see other user's ads, but will see their own
- users that have `adsDisabled=False` will get ads injected into their feed at query time
  - if the query to the user's feed already has one or more ad posts, no additional ads are added
  - if not, then an ad will be injected into the 2nd position in the returned set of posts (with at least one normal post before and after the ad. If there aren't enough normal posts in the response to do this, the ad is not injected) 
  - to select which ad to inject, each user has a pool of target `ad_feed` items that track how many times and when a user last saw an ad
  - the least-recently-seen ad in the users pool of target ads is injected into the feed
- payouts are done via the transactions service, although it is disabled for now until it can be tested
  - payouts are triggered by FOCUS views, with value of `Post.adPayment` and can happen at most once per `Post.adPaymentPeriod`
- implementation includes a new keys-only GSI on the main dynamo table, as well as a new secondary `ad_feed` dynamo table
  - user items are all added to that new GSI-K4 with a constant partition key, so we will be able to iterate over just user items in the future without doing a full table scan
- **IMPORTANT** includes a migration that must be run **manually** after the deployment is complete. The migration should be run before any ads are created.
- since this PR ads a GSI to the main table, deployment may take 'a while' on production. 10 minutes? 30 minutes? Looks like github actions timeout by default [after 360 minutes](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes) so we should be ok there :)


mfogel:
Closing this because github actions appear to be stuck using an old version of `.github/actions/ci.yml` which contains some broken caching which is causing the tests to fail. Will re-open shortly.

mfogel:
Resubmitted as #219

PR #227, opened 2021-06-09
Replace remaining dynamo transactions with async stream ops

mfogel:
This PR removes all remaining dynamo transactions from real-social. They were only still being used for chats.

Includes very few changes that would be visible to an external client (ie, the frontend) but here they are:

- automated system messages in group chats changed a bit (frontend doesn't really use group chats, AFAIK)
- a dating match now generates a group chat, instead of a direct chat, since two users can only have one direct chat and it's possible that two users already would have a direct chat open when they become a dating match


mfogel:
I will rebase this to deal with the conflicts

mfogel:
Ok, rebase done conflicts resolved. Ready for you when you have time @zkhin

zkhin (approved):

zkhin:
> Ok, rebase done conflicts resolved. Ready for you when you have time @zkhin

LGTM! Thanks @mfogel

PR #230, opened 2021-06-10
Add chat message error code

zkhin:
https://real-social.atlassian.net/browse/BE-168

### DIFF
--- a/real-main/app/handlers/appsync/handlers.py
+++ b/real-main/app/handlers/appsync/handlers.py
@@ -12,6 +12,7 @@ from app.models.appstore.exceptions import AppStoreException
 from app.models.block.enums import BlockStatus
 from app.models.block.exceptions import BlockException
 from app.models.card.exceptions import CardException
+from app.models.chat.error_codes import ChatErrorCode
 from app.models.chat.exceptions import ChatException
 from app.models.chat_message.enums import ChatMessageNotificationType
 from app.models.chat_message.exceptions import ChatMessageException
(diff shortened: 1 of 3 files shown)


real-main/app/handlers/appsync/handlers.py
```diff
@@ -1485,7 +1486,10 @@ def add_chat_message(caller_user, arguments, **kwargs):
 
     chat = chat_manager.get_chat(chat_id)
     if not chat or not chat.is_member(caller_user.id):
-        raise ClientException(f'User `{caller_user.id}` is not a member of chat `{chat_id}`')
+        raise ClientException(
+            f'User `{caller_user.id}` is not a member of chat `{chat_id}`',
+            [UserAuthError.USER_NOT_FOUND],
+        )
```
mfogel, line 1492: Not sure this error code makes sense. We don't know at this point if the user exists or not.

What do you think about this

```suggestion
if not chat:
    raise ClientException(
        'Chat `{chat.id}` does not exist',
        [ChatErrorCode.CHAT_NOT_FOUND],
    )

if not chat.is_member(caller_user.id):
    raise ClientException(
        'User `{caller_user.id}` is not a member of chat `{chat_id}`',
        [ChatErrorCode.NOT_A_MEMBER_OF_CHAT],
    )
```

As far as https://real-social.atlassian.net/browse/BE-168 goes, I think if the frontend were to display a message 'Chat does not exist' that would be good enough.

mfogel (changes requested):

mfogel (approved):

mfogel:
Hey @zkhin , when I run the new integration test in this PR `Cant send message to deleted user` in my dev env it doesn't pass, it fails. I get an error`ServerError: 'NoneType' object has no attribute 'id'`. Is it working for you?

